PR #3, opened 2018-10-17
First 10 solved

Hollister009:


tasks/9_find-numbers/find-numbers.js
```diff
@@ -0,0 +1,8 @@
+export const getNumbers = (str) => {
+  const reg = /\d+/g;
+  let strArr = str.match(reg);
+  let numArr = strArr.map((el) => {
```
olenatkachuk0501: let - is a signal that the variable may be reassigned) here both strArr and numArr are not reassignable

tasks/10_remove-whitespaces/remove-whitespaces.spec.js
```diff
@@ -1,11 +1,15 @@
-// import { trimWhiteSpaces } from './remove-whitespaces';
+import {trimWhiteSpaces} from './remove-whitespaces';
 
-xdescribe('remove-whitespaces', () => {
-    test('trimWhiteSpaces(‘I am travelling down the river’) should be equal to “Iamtravellingdowntheriver”', () => {
-        expect(trimWhiteSpaces('I am travelling down the river')).toBe('Iamtravellingdowntheriver');
-    });
+describe('remove-whitespaces', () => {
+  test('trimWhiteSpaces(‘I am travelling down the river’) should be equal to “Iamtravellingdowntheriver”', () => {
+    expect(trimWhiteSpaces('I am travelling down the river')).toBe(
```
olenatkachuk0501: you don't have to break a line until it fits)
P.S. Avoid lines longer than 80 characters) http://crockford.com/javascript/

tasks/2_convert-to-f/convert-to-f.js
```diff
@@ -0,0 +1,4 @@
+
+export const convertToF = (num) => {
+  return num * (9/5) + 32;
```
olenatkachuk0501: space between and after math symbols)

tasks/5_profile-lookup/profile-lookup.js
```diff
@@ -0,0 +1,36 @@
...
+    lastName: 'Vos',
+    number: 'unknown',
+    likes: ['Javascript', 'Gaming', 'Foxes'],
+  },
+];
+
+export const lookupProfile = (firstName, prop) => {
+  for (let i = 0; i < contacts.length; i++) {
+    if (contacts[i]['firstName'] === firstName) {
+      if (contacts[i].hasOwnProperty(prop)) return contacts[i][prop];
```
olenatkachuk0501: you can use in operator instead) https://developer.mozilla.org/uk/docs/Web/JavaScript/Reference/Operators/in

Hollister009: .hasOwnProperty() checks properties on object only. Instead of 'in' that checks it on Object.prototype also, which in my case is better. That's why I decided to choose this solution

tarasbobak:
Good job.

PR #5, opened 2018-10-17
[NATALIYA_MYKYTYUK] First 10 tasks

dnistrova:


tasks/3_is-boolean/is-boolean.js
```diff
@@ -0,0 +1,9 @@
+/* eslint "require-jsdoc": 0 */
+
+export function isBoolean(value) {
+  if (typeof(value) == typeof(true)) {
```
AlyonaPianykh: You can use ` 'boolean'` instead of typeof(true)
and 
It's enough to return just (typeof val == 'boolean') and not use if-else, because logical operator == is returning true or false

tasks/9_find-numbers/find-numbers.js
```diff
@@ -0,0 +1,6 @@
+/* eslint "require-jsdoc": 0 */
+
+export function getNumbers(str) {
+  str.match(/\d+/g).map(Number);
```
AlyonaPianykh: Where is `return` ?

tasks/3_is-boolean/is-boolean.js
AlyonaPianykh: Please rewrite it

tasks/5_profile-lookup/profile-lookup.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  {
+    firstName: 'Kristian',
+    lastName: 'Vos',
+    number: 'unknown',
+    likes: ['Javascript', 'Gaming', 'Foxes'],
+  },
+];
+
+export function lookupProfile(firstName, prop) {
```
AlyonaPianykh: Please create one more realization of this function using array methods, e.g. .forEach, .map etc.

PR #7, opened 2018-10-17
[STAS-MOLCHANOV]Tasks 1 10

moskdev:


tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,10 @@
+export function findLongestWord(str) {
+  let arr = str.split(/\W/);
```
tarasbobak: I like that you use regExp. This passes tests and is valid. But bear in mind this will not behave correctly for words with hyphen, like 'quick-thinking'.

Again no need to fix something here.

tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,10 @@
...
+  let maxLength =0;
+  for (let i = 0; i < arr.length; i++) {
```
tarasbobak: This is a usual case where you can do micro-optimizations.
Think of an array with [1,..., 999999999] items.
Then you are iterating over it and accessing it's length. Accessing length is quite costly operation.
So micro optimization would be:
const arrLength = arr.length;

for (let i = 0, i < arrLength; i++)

tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,10 @@
...
+    if (arr[i].length > maxLength) {
```
tarasbobak: Same idea can be applied here. You access arr[i].length at least once and when condition meets you access twice. Could be also a place to optimize. But that's not that important like the previous one.

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,7 @@
+export function titleCase(str) {
+  let arr = str.split(/\s/);
```
tarasbobak: easier way would be str.split(' ');

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,7 @@
...
+  for (let i = 0; i < arr.length; i++) {
+    arr[i] = arr[i].slice(0, 1).toUpperCase()+arr[i].slice(1).toLowerCase();
```
tarasbobak: arr[i].slice(0, 1) is OK. You can work with string similar to arrays. arr[i][0] is your first letter.

Bear in mind that logical and arithmetic operators should have whitespace before and after. I wonder how it passed eslint :)

tasks/2_convert-to-f/convert-to-f.js
```diff
@@ -0,0 +1,4 @@
+export function convertToF(celsius) {
+  let fahrenheit = 9 / 5 * celsius +32;
```
tarasbobak: spacing for +

is fahrenheit reassigned somewhere? use const instead for such cases

tasks/3_is-boolean/is-boolean.js
```diff
@@ -0,0 +1,3 @@
+export function isBoolean(value) {
+  return typeof(value) === 'boolean';
```
tarasbobak: it's written like typeof value not typeof(value)

tasks/5_profile-lookup/profile-lookup.js
```diff
@@ -0,0 +1,39 @@
...
+    'number': 'unknown',
+    'likes': ['Javascript', 'Gaming', 'Foxes'],
+  },
+];
+export const lookupProfile = function(fName, prop) {
+  let contactExists = false;
+  for (let i = 0; i < contacts.length; i++) {
+    if (contacts[i].firstName == fName) {
+      contactExists = true;
+      if (contacts[i].hasOwnProperty(prop)) {
```
tarasbobak: When if is in block like
```
if () {
...
} else {
...
}
```
else cannot be written the way yo do

tasks/5_profile-lookup/profile-lookup.js
```diff
@@ -0,0 +1,39 @@
...
+        return contacts[i][prop];
+      } else return 'No such property';
+    }
+  }
+  if (!contactExists) return 'No such contact';
+  return 'lul';
```
tarasbobak: LOL 👍

tasks/8_factorize-number/factorize-number.js
```diff
@@ -0,0 +1,9 @@
+export function factorialize(number) {
+  if (number == 0) return 1;
+  let result = 1;
+  while (number > 1) {
```
tarasbobak: I like this solution

tarasbobak:
Great job!
Please remove 11 and 12 tasks solution and commit them in new branch.

Besides of that please have a look at other way to iterate array and how they can be used: map, filter, reduce, forEach.

PR #8, opened 2018-10-18
Tasks 1 10

NFalinska:


tasks/3_is-boolean/is-boolean.js
```diff
@@ -0,0 +1,9 @@
+/* eslint "require-jsdoc": 0 */
+
+export function isBoolean(value) {
+  if (typeof value === 'boolean') {
```
tarasbobak: This one is quite important.

Instead of this one, write return typeof value === 'boolean' and think why it is much preferable

What you are basically doing is checking
```
if (true) {
  return true;
} else {
  return false;
}
```

tasks/5_profile-lookup/profile-lookup.js
```diff
@@ -0,0 +1,44 @@
...
+];
+
+export const lookupProfile = function(fName, prop) {
+  let contactExixst = false;
+  for (let i = 0; i<contacts.length; i++) {
+    if (contacts[i].firstName == fName) {
+      contactExixst = true;
+      if (contacts[i].hasOwnProperty(prop)) {
+        return contacts[i][prop];
+      } else return 'No such property';
```
tarasbobak: You cannot write it like that. The syntax is
```
if () {
  ...
} else {
  ....
}
```

An exception is if one-liner. It is OK for me, but I don't like it personally.

tarasbobak:
Good job. Please have a look at comments for future implementation.

tasks/1_golf-score/golf-score.js
```diff
@@ -0,0 +1,28 @@
...
+  EAGLE: 'Eagle',
+  BIRDIE: 'Birdie',
+  PAR: 'Par',
+  GO_HOME: 'Go Home!',
+  BOGEY: 'Bogey',
+  DOUBLE_BOGEY: 'Double Bogey',
+};
+export function golfScore(par, stroke) {
+  if (stroke === 1) {
+    return 'Hole-in-one!';
```
vovavoevidka: this should use constant instead, like this:
`return GOLF_SCORES.HOLE_IN_ONE;`

one more important think that you may be interested in, basically each section contains return
so else does not really needed!

it can be simplified like the following

```
if (stroke === 1) return GOLF_SCORES. HOLE_IN_ONE;
if (stroke <= par - 2 && stroke !== 1) return GOLF_SCORES. EAGLE;
if (stroke === par - 1) return GOLF_SCORES.BIRDIE;
etc.
```

tarasbobak:
@NFalinska Please pay attention for comments for future PR's.
Good job

PR #11, opened 2018-10-18
Tasks from 11 to 20

Hollister009:


tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,8 @@
+export const findLongestWord = (string) => {
+  const strArr = string.split(' ');
+  let longest = strArr[0].length;
+  strArr.forEach((el) => {
+    if (el.length >= longest) return (longest = el.length);
```
tarasbobak: this return doesn't make sense at all. This won't store the forEach loop.

tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,8 @@
+export const findLongestWord = (string) => {
+  const strArr = string.split(' ');
+  let longest = strArr[0].length;
```
tarasbobak: longest should be initialized with 0 in such case.

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,8 @@
+export const titleCase = (string) => {
+  const strArray = string.toLowerCase().split(' ');
+  let casedWordsArray = strArray.map((word) => {
```
tarasbobak: is casedWordsArray reassigned anywhere? If it is not use const.

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,8 @@
...
+    word = word.charAt(0).toUpperCase() + word.slice(1, word.length);
+    return word;
```
tarasbobak: just return word.charAt(0).toUpperCase() + word.slice(1, word.length);

There is no need to do any kind of reassignment

tasks/13_largest-of-four/largest-of-four.js
```diff
@@ -0,0 +1,7 @@
+export const largestOfFour = (array) => {
```
tarasbobak: You have implemented it in the way to pass tests. But that's not very good solution.

What if there were 5 subarrays with 4 items in each? for (let i = 3; i >= 0; i--) { won't work
You should have iterated over them

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,3 @@
+export const confirmEnding = (word, end) => {
+  return word.slice(-end.length) === end ? true : false;
```
tarasbobak: It is absolutely the same I wrote today on whiteboard.
return true ? true : false

You should just return word.slice(-end.length) === end

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,20 @@
+export const humanizeFormat = (num) => {
+  if (!num) return '';
+  const endings = ['st', 'nd', 'rd', 'th'];
+  let humanNum = num.toString();
+  switch (true) {
+    case humanNum.charAt(humanNum.length-1) === '1':
```
tarasbobak: Could this humanNum.charAt(humanNum.length - 1) be cached before the switch statement and then reused?

tasks/19_array-chunks/array-chunks.js
```diff
@@ -0,0 +1,16 @@
+export const arrayChunks = (array, size) => {
+  let twoDimensional = [];
+  let n = 0;
+  let i = array.length;
+  while (i - size > 0) {
+    i -= size;
+    twoDimensional.push(
+        array
+            .slice(n, n + size)
+            .reduce((sum, cur) => sum.concat(cur), [])
```
tarasbobak: interesting solution, but I don't understand why do you need this reduce?

PR #14, opened 2018-10-18
Task 12 20

art1kos:
10 tasks from 12 to 20 with 5th task


tasks/10_remove-whitespaces/remove-whitespaces.js
```diff
@@ -0,0 +1,4 @@
+export const trimWhiteSpaces = (str) =>{
+  let result = str.split(' ').join('');
```
tarasbobak: Use const when you will not reassign it.

An basically you don't need this variable here. You can just return

tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,10 @@
+export const findLongestWord = (str) =>{
+  let words = str.split(' ');
```
tarasbobak: const when you will not reassign

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,8 @@
+export const titleCase = (str) => {
+  let newStr = str.split(' ');
+  for ( let i = 0; i < newStr.length; i++ ) {
+    let upperLetter = newStr[i].charAt(0).toUpperCase();
```
tarasbobak: You can also deal with string same way as with arrays. Like 'HELLO'[0] will be H
But charAt is also fine!

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,8 @@
+export const titleCase = (str) => {
+  let newStr = str.split(' ');
+  for ( let i = 0; i < newStr.length; i++ ) {
```
tarasbobak: No whitespaces between these braces. for (let i ...)

tasks/14_clone-arr/clone-arr.js
```diff
@@ -0,0 +1,3 @@
+export const cloneArr = (arr) =>{
```
tarasbobak: Good. If arrow function just returns and is very small it can be on-liner
export const cloneArr = arr => arr.slice();

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,10 @@
+export const confirmEnding = (str, el) => {
+  let res = false;
```
tarasbobak: do you need res at all???

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,10 @@
...
+  let elLength = el.length;
+  let strLength = str.length;
+  let ending = str.substring(strLength - elLength, str.length);
```
tarasbobak: why no reuse strLength as second param?

tasks/17_truncate-string/truncate-string.js
```diff
@@ -0,0 +1,5 @@
+export const truncateString= (str, n) =>{
+  if (n>=str.length) return str;
```
tarasbobak: Watch proper spacing if (condition >= secondCondition)

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,13 @@
+export const humanizeFormat = (num) => {
+  if (!(num)) {
```
tarasbobak: You can cache ending in a var and reuse instead of duplicating operation

tasks/19_array-chunks/array-chunks.js
```diff
@@ -0,0 +1,10 @@
+export const arrayChunks = (arr, num) => {
+  let newArr = arr.slice();
+  let res = [];
+  for (let i = 0; i < Math.ceil(arr.length / num); i++) {
```
tarasbobak: Math.ceil(arr.length / num) can be cached in a variable to not evaluate it in every loop, but once.

tasks/19_array-chunks/array-chunks.js
```diff
@@ -0,0 +1,10 @@
...
+    let arrChunk = [];
```
tarasbobak: why is this initialized with empty array?

PR #16, opened 2018-10-18
[Sophia]Tasks 11 to 20

SofiMat:


tarasbobak:
Mess is a great name for commit :)

vovavoevidka:
Please do not use `/* eslint-disable */ `
It disables entire file

SofiMat:
i removed it from my files

пт, 19 жовт. 2018 о 16:01 Volodymyr Voyevidka <notifications@github.com>
пише:

tasks/14_clone-arr/clone-arr.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable */ 
+
+export function cloneArr(arr) {
+    let clone = arr.slice(0);
+    return clone;
```
vovavoevidka: `return arr.slice(0);`
you do not need extra variable here

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable */ 
+
+export function confirmEnding (str, target) {
+    let reg = new RegExp(target);
+    let nstr=str.split(' ');
```
vovavoevidka: i don't think you need this str=>arr conversion...

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,11 @@
...
+    if (nstr[nstr.length-1] === target || target === str.substring(Number(str.length-1), Number(str.length)) || str.slice(-target.length)=== target) {
```
vovavoevidka: this is similar to
```
if (<expressions>) {
  return true;
} else {
  return false;
}

```

why not just `return <expressions>;`?

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,27 @@
...
+        } 
+        if (str[str_len] == 3) {
+            return str+'rd';
+        } 
+        if (str[str_len] >= 4){
+            return str+'th';
+        } else {
+            return final;
+        }}
+    else {
```
vovavoevidka: you do not need `else` when in every if condition you are doing `return`
you can simply add `return` in the end of the function

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable */ 
+
+export function humanizeFormat(num) {
+    let final = '';
+    if(num){
+        let str = num.toString();
+        let str_len = str.length-1;
+        if(str[str_len]==1) {
```
vovavoevidka: you can use switch here instead.

tasks/19_array-chunks/array-chunks.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */ 
+
+export function arrayChunks (arr, size) {
+    let res= [];
+    while (arr.length) {
```
vovavoevidka: nice!

tasks/20_truncate-array/truncate-array.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable */ 
+
+export function truncateArray (arr ,n) {
+    let newArr=[];
```
vovavoevidka: you do not need this extra variable

tasks/20_truncate-array/truncate-array.js
```diff
@@ -0,0 +1,11 @@
...
+    if(n>0){
+        newArr = arr.splice(n, arr.length-1);
```
vovavoevidka: `return arr.splice(n, arr.length-1);` is enough

tasks/13_largest-of-four/largest-of-four.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable */ 
+
+export function largestOfFour(arr) {
```
vovavoevidka: @tarasbobak I have a concern about this implementation!
It work with existing set of tests but it has wrong logic
For instance this test will fail:
```
expect(largestOfFour([
            [4, 9, 1, 3],
            [13, 35, 18, 26],
            [32, 35, 97, 39],
            [1000000, 1001, 857, 1],
            [1000000, 1001, 857, 1],
            [1000000, 1001, 857, 1]
        ])).toEqual([9, 35, 97, 1000000, 1000000, 1000000]);
```

Please suggest on this

tarasbobak: I'm not sure if it will fail. But basically you don't need inner for loop here.
```
for(let i = 0; i < arr.length; i++) {
  arrMax.push(Math.max.apply(null, arr[i]));
}
```

SofiMat:
My fault, will change it in 10 minutes

SofiMat:
> @SofiMat You did not remove it from your files or did not push your change. I can see them in the code.

fixed for 11-20 tasks

tasks/11_find-longest-word/find-longest-word.js
```diff
@@ -0,0 +1,10 @@
+export function findLongestWord(str) {
+  let arr = str.split(' ');
+  let max = 0;
+  let lengArr = [];
+  for (let i=0; i < arr.length; ++i) {
```
tarasbobak: A better solution is to have maxLength variable an in each iteration of for loop compare each word to max. Thus you would not need to apply Math.max(...). Under the hood it again loops over items to check for Max value

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,10 @@
+
+export function confirmEnding(str, target) {
+  let nstr=str.split(' ');
+  /* eslint-disable-next-line */
+  if (nstr[nstr.length-1] === target || target === str.substring(Number(str.length-1), Number(str.length)) || str.slice(-target.length)=== target) {
```
tarasbobak: This is a total mess :) What is the difference between nstr[nstr.length-1] === target and str.slice(-target.length)=== target. 
This part seems to be the same as previous ones target === str.substring(Number(str.length-1), Number(str.length))

tarasbobak: if (cond) {
  return true;
} else {
  return false;
}

just return cond;

tasks/16_repeat-string/repeat-string.js
```diff
@@ -0,0 +1,8 @@
+
+export function repeatStringNumTimes(str, num) {
+  if (num > 0) {
```
tarasbobak: Please take into account previous Volodymyr's comment.
When you return in if statement, you don't need else expicit block.
```
  if (num > 0) {
    return str.repeat(num);
  }
  return '';
```

tasks/17_truncate-string/truncate-string.js
```diff
@@ -0,0 +1,12 @@
+
+export const STRING0 = 'A-tisket a-tasket A green and yellow basket';
```
tarasbobak: Why have you added those consts here ?)

SofiMat:
Oh, I thought I need to

сб, 20 жовт. 2018 о 22:37 Taras <notifications@github.com> пише:

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,26 @@
+
+export function humanizeFormat(num) {
+  let final = '';
+  if (num) {
+    let str = num.toString();
+    let strLen = str.length-1;
```
tarasbobak: not a good name. srtLen is a good name for str.length but not for str.length - 1

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,26 @@
...
+    if (str[strLen]==1) {
+      return str+'st';
+    }
+    if (str[strLen] == 2) {
+      return str+'nd';
```
tarasbobak: return str[space]+[space]'nd';

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,26 @@
+
+export function humanizeFormat(num) {
+  let final = '';
+  if (num) {
+    let str = num.toString();
+    let strLen = str.length-1;
+    if (str[strLen]==1) {
```
tarasbobak: Why not ===

SofiMat:
Formatting? But eslint let me do it

сб, 20 жовт. 2018 о 22:38 Taras <notifications@github.com> пише:

tasks/19_array-chunks/array-chunks.js
```diff
@@ -0,0 +1,7 @@
+export function arrayChunks(arr, size) {
+  let res= [];
+  while (arr.length) {
```
tarasbobak: That's pretty interesting solution!
But generally saying it's not a good practice to change incoming data. What I mean is that you you splice which changes arr argument itself.

tasks/20_truncate-array/truncate-array.js
```diff
@@ -0,0 +1,10 @@
+
+export function truncateArray(arr, n) {
+  let newArr=[];
+  if (n>0) {
```
tarasbobak: @SofiMat Please pay more attention to proper indentation and whitespace. It may seem not serious from the beginning, but it is very important!

tasks/20_truncate-array/truncate-array.js
```diff
@@ -0,0 +1,10 @@
...
+    newArr = arr.splice(n, arr.length-1);
```
tarasbobak: please omit unneeded variables, like newArr in that case

tarasbobak: you could implement an easier solution without modifying incoming array.

SofiMat:
Sure, I will. In next pull requests

сб, 20 жовт. 2018 о 22:42 Taras <notifications@github.com> пише:

SofiMat:
So, let newArr = arr.splice(n, arr.length-1); is enough?


сб, 20 жовт. 2018 о 22:44 Taras <notifications@github.com> пише:

tarasbobak:
@SofiMat Thank you for your work. 👍 
For future:
Pay more attention to the way you write code. Code should be well-formatted and easy to read.
When you finish task and it works ask yourself a question if there are better solution for that task.
Try not to modify incoming data in your future implementation.

SofiMat:
Thanks for the comments, I’ll try to improve myself (though tasks 21-30 and
31-40 may have same remarks, cause I pushed them before I got any comments,
but I’ll fix this henceforth)

сб, 20 жовт. 2018 о 22:49 Taras <notifications@github.com> пише:

PR #17, opened 2018-10-18
[NATALIYA_MYKYTYUK] Tasks 11-20

dnistrova:


tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,9 @@
+export function titleCase(str) {
+  str = str.toLowerCase().split(' ');
```
Nazar-bnm: The best way use a pure function. Means that input argument isn't remapped.

tasks/14_clone-arr/clone-arr.js
```diff
@@ -0,0 +1,4 @@
+export function cloneArr(arr) {
+  let cloneArr = arr.slice();
```
Nazar-bnm: Use const where variable won't reassign.

tasks/16_repeat-string/repeat-string.js
```diff
@@ -0,0 +1,5 @@
+export function repeatStringNumTimes(string, num) {
+  if (num < 0) return '';
+  if (num === 1) return string;
+  else return string + repeatStringNumTimes(string, num - 1);
```
Nazar-bnm: use curly brackets for if-else construction for readability code

PR #18, opened 2018-10-18
ydemkiv tasks_11_20

YarynaD:


tasks/13_largest-of-four/largest-of-four.js
```diff
@@ -0,0 +1,12 @@
+/* eslint "require-jsdoc": 0 */
+
+export function largestOfFour(mainArray) {
+  return mainArray.map(function(subArray) {
+    return subArray.reduce(function(previousLargestNumber,
```
AlyonaPianykh: This task supposed to be implemented using `for` loop. But your implementation is ok.

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,9 @@
+/* eslint "require-jsdoc": 0 */
+
+export function confirmEnding(string, target) {
+  if (string.substr(-target.length) === target) {
+    return true;
```
AlyonaPianykh: nice, but you can write just `string.substr(-target.length) === target` and it would return needed boolean

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,16 @@
+export function humanizeFormat(number) {
+  if (typeof(number) == 'undefined') return '';
+  number = number.toString();
+  let ending = 'th';
+  if (number.slice(-1) == 1) {
```
AlyonaPianykh: You can use `switch` for this task

PR #20, opened 2018-10-19
[pavlo_turchyn]Task 11 20

FabianMoretti:


prezlyataa:
You have conflicts here, should resolve it.

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,13 @@
+/* eslint "require-jsdoc": 0 */
+
+export const titleCase = (str) => {
+  let lowStr = str.toLowerCase();
```
tarasbobak: You need to understand the difference between let and const. Const is used for values that will not be reassigned. Let on the other hand is used for variables which will be reassigned

tasks/12_title-case/title-case.js
```diff
@@ -0,0 +1,13 @@
...
+  let lowArr = lowStr.split(' ');
+  let resArr = [];
+  let lowArrLength = lowArr.length;
+  for (let i = 0; i < lowArrLength; i++) {
+    resArr.push(lowArr[i][0].toUpperCase() +
```
tarasbobak: I don't like the way you wrapped this line
You could go one line:
resArr.push(lowArr[i][0].toUpperCase() + lowArr[i].slice(1));
Or make variable for first uppercase value and reused it
Please read about line wrapping rules:
https://google.github.io/styleguide/jsguide.html#formatting-line-wrapping

tasks/13_largest-of-four/largest-of-four.js
```diff
@@ -0,0 +1,9 @@
+/* eslint "require-jsdoc": 0 */
+
+export const largestOfFour = (bigArr) => {
+  let maxArr = [];
+  for (let i = 0; i < bigArr.length; i++) {
+    maxArr.push(Math.max.apply(null, bigArr[i]));
```
tarasbobak: nice

tasks/14_clone-arr/clone-arr.js
```diff
@@ -0,0 +1,5 @@
+/* eslint "require-jsdoc": 0 */
+
+export const cloneArr = (original) => {
+  return original.concat();
```
tarasbobak: interesting 👍

tasks/15_confirm-ending/confirm-ending.js
```diff
@@ -0,0 +1,5 @@
+/* eslint "require-jsdoc": 0 */
+
+export const confirmEnding = (str, target) => {
+  return str.substr(-target.length)===target;
```
tarasbobak: add whitespace between ===, like `return str.substr(-target.length) === target;`

tasks/16_repeat-string/repeat-string.js
```diff
@@ -0,0 +1,9 @@
+/* eslint "require-jsdoc": 0 */
+
+export const repeatStringNumTimes = (str, counter) => {
+  let res = '';
+  for (let i = 0; i < counter; i++) {
+    res+=str;
```
tarasbobak: res += str;

tasks/17_truncate-string/truncate-string.js
```diff
@@ -0,0 +1,7 @@
+/* eslint "require-jsdoc": 0 */
+
+export const truncateString = (str, n) => {
+  if (n>=str.length) return str;
```
tarasbobak: if (n >= str.length)
if (n > 3)

tasks/18_humanize-format/humanize-format.js
```diff
@@ -0,0 +1,11 @@
+/* eslint "require-jsdoc": 0 */
+
+export const humanizeFormat = (num) => {
+  let strNum = num + '';
+  let lastStrNum =strNum[strNum.length - 1];
```
tarasbobak: it's good that you cached value, but it has to be const first.
Secondly you could make number out of string once and not each time, like:
const ending = parseInt(strNum[strNum.length - 1], 10);

tasks/19_array-chunks/array-chunks.js
```diff
@@ -0,0 +1,11 @@
+/* eslint "require-jsdoc": 0 */
+
+export const arrayChunks = (mainArr, size) => {
+  let result = [];
+  let i = 0;
+  let n = mainArr.length;
```
tarasbobak: n does not tell a lot to a reader. Consider better naming. like arrLength

tasks/20_truncate-array/truncate-array.js
```diff
@@ -0,0 +1,5 @@
+/* eslint "require-jsdoc": 0 */
+export const truncateArray = (arr, size) => {
+  arr.splice(0, size);
```
tarasbobak: interesting, but generally it is not a good practice to change incoming data. So ideally you should not change the value for arr itself but return a new array.

tarasbobak:
Good job. Please take into account comments for future work!

PR #24, opened 2018-10-19
[ANDRIY_ZHUKEVYCH] Fixed tasks 11-20

Hollister009:


tasks/13_largest-of-four/largest-of-four.js
```diff
@@ -1,7 +1,8 @@
 export const largestOfFour = (array) => {
   let maxNumArr = [];
-  for (let i = 3; i >= 0; i--) {
-    maxNumArr.unshift(Math.max(...array[i]));
+  let arrIndex = array.length - 1;
+  for (; arrIndex >= 0; arrIndex--) {
```
tarasbobak: first part of for loop will be executed only once, so there is no need to declare it before for loop

PR #25, opened 2018-10-19
[ANDRIY_ZHUKEVYCH]Tasks 21-30

Hollister009:


tasks/30_people-in-bus/people-in-bus.js
```diff
@@ -0,0 +1,8 @@
+export const peopleInBus = (bus) => {
+  let peopleInBus = 0;
+  bus.forEach((stop) => {
```
AlyonaPianykh: Also you can use destructuring from ES6 to make code look better

```
bus.forEach(( [getIn, getOut]) => {
    inBus += getIn;
    inBus -= getOut;
...
```
See the link: `https://dev.to/sarah_chima/destructuring-assignment---arrays-16f`

tarasbobak: why not a single peopleInBus = peopleInBus + stop[0] - stop[1]?

tasks/21_check-letters/check-letters.js
```diff
@@ -0,0 +1,10 @@
+export const checkLetters = (arr) => {
+  let flag = true;
+  const word = arr[0].toLowerCase();
+  const letters = arr[1].toLowerCase().split('');
+  letters.every((letter) => {
```
tarasbobak: Wrong use of every.

tarasbobak: Every itself returns true or false in the end
In your code you don't need to store flag in variable at all

tasks/24_index-to-insert/index-to-insert.js
```diff
@@ -0,0 +1,6 @@
+export const getIndexToInsert = (array, element) => {
+  let modifiedArr = array.slice();
```
tarasbobak: This solution is inefficient.
Array is already sorted.
You generally have to iterate over array and find first el which is bigger than the one you want to insert. And then insert before it.

Hollister009: How this is sorted ?
getIndexToInsert([2, 20, 10], 19)
getIndexToInsert([3, 10, 5], 3)

tarasbobak:
Good job!

PR #27, opened 2018-10-19
[NATALIYA_MYKYTYUK] Tasks 21-30

dnistrova:


tasks/26_average-mark/average-mark.js
```diff
@@ -0,0 +1,5 @@
+export function getAverage(arr) {
+  let sum = arr.reduce((previous, current) => current += previous);
```
tarasbobak: const

tasks/29_remove-duplicates/remove-duplicates.js
```diff
@@ -0,0 +1,10 @@
+export function removeDuplicates(arr) {
+  let uniqueArray = [];
```
tarasbobak: Solution is good, use const

tasks/30_people-in-bus/people-in-bus.js
```diff
@@ -0,0 +1,13 @@
+export function peopleInBus(busStops) {
+  let totalPeople = 0;
+  for (let i=0; i<busStops.length; i++) {
+    let currentStop = busStops[i];
+    let getOn = currentStop[0];
+    let getOff = currentStop[1];
+
+    totalPeople += getOn;
```
tarasbobak: why not do addition and subtraction in a single statement?

tasks/25_rot13-encoder/rot13-encoder.js
```diff
@@ -0,0 +1,14 @@
+export function rot13Encoder(str) {
+  let arr = [];
```
tarasbobak: Great solution, but use const if there will be no reassignment

tasks/22_remove-falsy-values/remove-falsy-values.js
```diff
@@ -0,0 +1,7 @@
+export function removeFalsyValues(arr) {
+  return arr.filter(function(value) {
+    if (value) {
+      return (value);
```
tarasbobak: just return value, it will be transformed to boolean

tasks/21_check-letters/check-letters.js
```diff
@@ -0,0 +1,11 @@
+export function checkLetters(arr) {
+  arr[0] = arr[0].toLowerCase();
+  arr[1] = arr[1].toLowerCase();
+  for (let i = 0; i < arr[1].length; i++) {
+    if (arr[0].indexOf(arr[1][i]) < 0) {
```
tarasbobak: When item is not found it is -1. So you should explicitly check for -1, not <0

PR #29, opened 2018-10-19
[Artur Kostiuk] Task 21 30

art1kos:
tasks from 21 - 30

### DIFF
--- a/tasks/21_check-letters/check-letters.js
+++ b/tasks/21_check-letters/check-letters.js
@@ -1,0 +1,11 @@
+export const checkLetters = (arr) => {
+  const arr1New = arr[0].toLowerCase().split('');
+  const arr2New = arr[1].toLowerCase().split('');
+  const arr2NewLength = arr2New.length;
+  for (let i = 0; i < arr2NewLength; i++) {
+    if (arr1New.indexOf(arr2New[i]) == -1) {
+      return false;
+    }
+  }
+  return true;
+};
--- a/tasks/23_seek-destroy/seek-destroy.js
+++ b/tasks/23_seek-destroy/seek-destroy.js
@@ -1,12 +1,7 @@
-export const destroyer = (arr, ...args) => {
-  const arrLength = arr.length;
-  const argsLength = args.length;
-  for (let i = 0; i < arrLength; i++) {
-    for (let j = 0; j < argsLength; j++) {
-      if (arr[i] === args[j]) {
-        delete arr[i];
-      }
-    }
+export const destroyer = (arr, ...args ) => {
+  return arr.filter(destroyNum);
+
+  function destroyNum(element) {
+    return !args.includes(element);
   }
-  return arr.filter(Boolean);
 };
--- a/tasks/23_seek-destroy/seek-destroy.js
+++ b/tasks/23_seek-destroy/seek-destroy.js
@@ -1,0 +1,12 @@
+export const destroyer = (arr, ...args) => {
+  const arrLength = arr.length;
+  const argsLength = args.length;
+  for (let i = 0; i < arrLength; i++) {
+    for (let j = 0; j < argsLength; j++) {
+      if (arr[i] === args[j]) {
+        delete arr[i];
+      }
+    }
+  }
+  return arr.filter(Boolean);
+};
--- a/tasks/25_rot13-encoder/rot13-encoder.js
+++ b/tasks/25_rot13-encoder/rot13-encoder.js
@@ -1,0 +1,21 @@
+export const rot13Encoder = (str) => {
+  const alphabet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
+  const rot13Alphabet = 'NOPQRSTUVWXYZABCDEFGHIJKLM'.split('');
+  let result = '';
+  const strLength = str.length;
+  const alphabetLength = alphabet.length;
+  for (let i = 0; i < strLength; i++) {
+    for (let j = 0; j < alphabetLength; j++) {
+      if (str[i] === alphabet[j]) {
+        result += rot13Alphabet[j];
+      }
+    }
+    if (str[i] === ' ') {
+      result += ' ';
+      continue;
+    } else if (!str[i].match(/[A-Z]/)) {
+      result += str[i];
+    }
+  }
+  return result;
+};
--- a/tasks/26_average-mark/average-mark.js
+++ b/tasks/26_average-mark/average-mark.js
@@ -1,0 +1,9 @@
+export const getAverage = (arr) => {
+  let sumMarks = 0;
+  const markLength = arr.length;
+  for (let i = 0; i < markLength; i++) {
+    sumMarks += arr[i];
+  }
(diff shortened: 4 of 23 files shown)


tasks/23_seek-destroy/seek-destroy.js
```diff
@@ -0,0 +1,12 @@
...
+  const arrLength = arr.length;
+  const argsLength = args.length;
+  for (let i = 0; i < arrLength; i++) {
+    for (let j = 0; j < argsLength; j++) {
+      if (arr[i] === args[j]) {
+        delete arr[i];
+      }
+    }
+  }
+  return arr.filter(Boolean);
```
prezlyataa: You delete all values that equals false, when you pass such arguments destroyer([2, 3, 2, 3, 0, 5], 2, 3) function will return [5], but result must be [0, 5]. In your case 0 == false and this is true, but you should return 0, because this is number too. You should rewrite it.

tasks/28_vowel-count/vowel-count.js
```diff
@@ -0,0 +1,13 @@
+export const getVowelCount = (str) => {
+  str = str.toLowerCase();
```
prezlyataa: You don't need use here additional step, it will be better write in one string 
`const newStr = str.toLowerCase().split('');`
Also for variable vowels use const.

prezlyataa: You don't need use here additional step, it will be better write in one string 
`const newStr = str.toLowerCase().split('');`
Also for variable vowels use const.

prezlyataa: You don't need use here additional step, it will be better write in one string 
`const newStr = str.toLowerCase().split('');`
Also for variable vowels use const.

prezlyataa: You don't need use here additional step, it will be better write in one string 
`const newStr = str.toLowerCase().split('');`
Also for variable vowels use const.

tasks/23_seek-destroy/seek-destroy.js
prezlyataa: It will be fail when you pass such arguments like
`destroyer([2,3,1,0],2,3)`
You delete all values that equal false and also in this case 0 == false it will be true, but you should return 0, this is number too.
You should rewrite it.

tasks/29_remove-duplicates/remove-duplicates.js
```diff
@@ -0,0 +1,5 @@
+export const removeDuplicates = (arr) => {
+  /* eslint-disable-next-line */
+  let res =  [...new Set(arr)];
```
prezlyataa: You don't need declare variable here, just `return [...new Set(arr)];`

tasks/21_check-letters/check-letters.js
```diff
@@ -0,0 +1,11 @@
+export const checkLetters = (arr) => {
+  const arr1New = arr[0].toLowerCase().split('');
+  const arr2New = arr[1].toLowerCase().split('');
+  const arr2NewLength = arr2New.length;
+  for (let i = 0; i < arr2NewLength; i++) {
+    if (arr1New.indexOf(arr2New[i]) == -1) {
```
tarasbobak: use strict comparison by default ===

tasks/21_check-letters/check-letters.js
```diff
@@ -0,0 +1,11 @@
+export const checkLetters = (arr) => {
+  const arr1New = arr[0].toLowerCase().split('');
+  const arr2New = arr[1].toLowerCase().split('');
```
tarasbobak: I would say arr1New and arr2New is not very good naming

tarasbobak: I would say arr1New and arr2New is not very good naming

tarasbobak: I would say arr1New and arr2New is not very good naming

tarasbobak: I would say arr1New and arr2New is not very good naming

tasks/23_seek-destroy/seek-destroy.js
tarasbobak: I see this second time :) It is not a good solution.

tarasbobak: I see this second time :) It is not a good solution.
args is an array so you can use indexOf to check if it contains certain value.
Not need for an inner loop at all. It is unefficient

tasks/25_rot13-encoder/rot13-encoder.js
```diff
@@ -0,0 +1,21 @@
+export const rot13Encoder = (str) => {
+  const alphabet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
+  const rot13Alphabet = 'NOPQRSTUVWXYZABCDEFGHIJKLM'.split('');
+  let result = '';
+  const strLength = str.length;
+  const alphabetLength = alphabet.length;
+  for (let i = 0; i < strLength; i++) {
+    for (let j = 0; j < alphabetLength; j++) {
```
tarasbobak: It is not very efficient. The same way you hardcoded strings you could calculate it when needed.

tasks/25_rot13-encoder/rot13-encoder.js
```diff
@@ -0,0 +1,21 @@
...
+      if (str[i] === alphabet[j]) {
+        result += rot13Alphabet[j];
+      }
+    }
+    if (str[i] === ' ') {
```
tarasbobak: This does not make sense to me.
If I'm not mistaken this condition
```
if (!str[i].match(/[A-Z]/)) {
      result += str[i];
}
```
will work for whitespace too. So first if can be handled by your else if.

tasks/26_average-mark/average-mark.js
```diff
@@ -0,0 +1,9 @@
+export const getAverage = (arr) => {
+  let sumMarks = 0;
```
tarasbobak: That whole logic could be handled by a single reduce. But it works and seems fine

tasks/28_vowel-count/vowel-count.js
```diff
@@ -0,0 +1,13 @@
...
+  const newStr = str.split('');
+  const newStrLength = newStr.length;
+  let vowels = ['a', 'e', 'i', 'o', 'u'];
+  let vowelsCount = 0;
+  for (let i = 0; i < newStrLength; i++) {
```
tarasbobak: This solution is also fine. But reduce here would be less verbose.

tasks/30_people-in-bus/people-in-bus.js
```diff
@@ -0,0 +1,9 @@
+export const peopleInBus = (arr) => {
+  let resIn = arr.reduce((a, c) => {
```
tarasbobak: const when variable will not be reassigned.
Solution is interesting and could be done with a single reduce.

tarasbobak:
@art1kos Please rewrite seek-destroy.

PR #30, opened 2018-10-19
[Y_Demkiv]Tasks_21_30

YarynaD:


tasks/24_index-to-insert/index-to-insert.js
```diff
@@ -0,0 +1,7 @@
+export function getIndexToInsert(arr, num) {
+  arr.push(num);
+  arr.sort(function(a, b) {
```
IgorStetsiuk: you can write it such way `.sort((a, b) => a - b)`

IgorStetsiuk: you can write it such way `.sort((a, b) => a - b)`

IgorStetsiuk: you can write it such way `.sort((a, b) => a - b)`

tasks/23_seek-destroy/seek-destroy.js
```diff
@@ -0,0 +1,10 @@
+export function destroyer(arr, ...args) {
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < args.length; j++) {
```
tarasbobak: Args is an array so you can use includes or indexOf to check for presence of a value. Second loop will be not needed

tasks/25_rot13-encoder/rot13-encoder.js
```diff
@@ -0,0 +1,12 @@
+export function rot13Encoder(str) {
+  return str.split('')
+      .map.call(str, function(char) {
```
tarasbobak: This solution is very interesting! But you don't need call here

tasks/26_average-mark/average-mark.js
```diff
@@ -0,0 +1,3 @@
+export function getAverage(marks) {
```
tarasbobak: 👍

tasks/30_people-in-bus/people-in-bus.js
```diff
@@ -0,0 +1,10 @@
+export function peopleInBus(Stops) {
+  let Summ = 0;
```
tarasbobak: do not uppercase variables. It is reserved for classed. Use only camelCase

tarasbobak:
Overly good job. Use const by default. Do not use let until you know it will be reassigned.

PR #31, opened 2018-10-20
[Sophia]Tasks 31 to 40

SofiMat:


tasks/31_reverse-string-in-range/reverse-string-in-range.js
```diff
@@ -0,0 +1,8 @@
+export function reverseStringInRange(str, arr) {
+  let copyStr = str.split('');
+  let chunk = copyStr.slice(arr[0], arr[1]+1).reverse().join('');
+  copyStr.splice(arr[0], chunk.length, chunk);
+  let final = copyStr.join('');
```
tarasbobak: const as default for non reassignable values
final is not needed, just return the result of this computation

tasks/32_numbers-to-reversed-array/numbers-to-reversed-array.js
```diff
@@ -0,0 +1,10 @@
+export function numberToReversedArray(num) {
+  let str = num.toString();
```
tarasbobak: This could be done in a single command, something like
return num.toString().split('').reverse().map(Number);

tasks/33_array-plus-array/array-plus-array.js
```diff
@@ -0,0 +1,21 @@
+export function arrayPlusArray(arr1, arr2) {
+  let total = [];
+  const leng1 = arr1.length;
+  const leng2 = arr2.length;
+  if (leng1 === leng2) {
+    for ( let i = 0; i < leng1; i++) {
+      total.push(arr1[i]+arr2[i]);
```
tarasbobak: arr1[i]{space}+{space}arr2[i]

tasks/33_array-plus-array/array-plus-array.js
```diff
@@ -0,0 +1,21 @@
...
+    }
+  } else if (leng1<leng2) {
```
tarasbobak: This solution seems pretty complicated :)
const result = copy of first array;
iterate second array
    if value exists in array at certain position sum them
    if not set value

return

tasks/34_max-triplet-sum/max-triplet-sum.js
```diff
@@ -0,0 +1,10 @@
+export function maxTripletSum(arr) {
+  /* eslint-disable-next-line */
+  let newArr = [...new Set(arr)];
+  let sorted = newArr.sort( function(a, b) {
+    return b - a;
+  });
+  let sum = 0;
+  sum = sorted[0] + sorted[1] + sorted[2];
+  return sum;
```
tarasbobak: just return the sum, don't introduce new vars

tasks/35_array-leaders/array-leaders.js
```diff
@@ -0,0 +1,18 @@
+export function arrayLeaders(arr) {
+  let newArr = [];
```
tarasbobak: newArr is never reassigned -> const

tasks/36_row-weight/row-weight.js
```diff
@@ -0,0 +1,25 @@
...
+  for (let i=0; i<leng; ++i) {
+    if (i%2 !== 0) {
+      team2.push(arr[i]);
+    } else {
+      team1.push(arr[i]);
+    }
+  }
+  const final = [];
+  if (arr.length !== 1) {
+    let num1 = team1.reduce(function(sum, current) {
```
tarasbobak: Start with 0 for both reduce and you won't need else statement. If array will be empty it will be 0.

tarasbobak:
No, we are talking about reassignment.
Like arr1 = [1,2,3];
arr1 = [4,5,6]

tasks/37_min-number-to-prime/min-number-to-prime.js
```diff
@@ -0,0 +1,33 @@
...
+    let num = n + 1;
+    while (num) {
+      if (isPrime(num)) {
+        return num;
+      }
+      num++;
+    }
+    return 0;
+  }
+  let sum = 0;
```
tarasbobak: calculate sum with reduce in future

tasks/37_min-number-to-prime/min-number-to-prime.js
```diff
@@ -0,0 +1,33 @@
...
+    for (let i = 2; i < n; i++) {
+      if (n % i === 0) {
+        return false;
+      }
+    }
+    return true;
+  }
+  function findPrime(n) {
+    let num = n + 1;
+    while (num) {
```
tarasbobak: here you are relying on the return statement. Because if function will not return in while statement you loop will be going forever

tasks/38_max-product/max-product.js
```diff
@@ -0,0 +1,11 @@
+export function maxProduct(inputArray) {
```
tarasbobak: Solution is valid. <, *, + needs proper spacing

PR #37, opened 2018-10-22
[STAS-MOLCHANOV] tasks 21-30

moskdev:


tasks/24_index-to-insert/index-to-insert.js
```diff
@@ -0,0 +1,8 @@
+export function getIndexToInsert(arr, value) {
+  arr.push(value);
+  const sortArr = arr;
```
Nazar-bnm: this assignment unnecessary, because array (object) is copied by link, so input array is modified too.

tasks/23_seek-destroy/seek-destroy.js
```diff
@@ -0,0 +1,5 @@
+export function destroyer(arr, ...theArgs) {
+  return arr.filter((element) => {
+    return (theArgs.indexOf(element) == -1);
```
tarasbobak: Always use strict comparisons === and !==

tasks/28_vowel-count/vowel-count.js
```diff
@@ -0,0 +1,8 @@
+export function getVowelCount(str) {
```
tarasbobak: great!

tasks/30_people-in-bus/people-in-bus.js
```diff
@@ -0,0 +1,7 @@
+export function peopleInBus(arr) {
```
tarasbobak: cool

PR #38, opened 2018-10-22
[TARAS-YAKYMUIK]41-50

taras1313:


tasks/42_first-n-smallest/first-n-smallest.js
```diff
@@ -0,0 +1,11 @@
+/* eslint "require-jsdoc": 0 */
+
+export function firstNSmallest(arr, n) {
+  for (let i = 0; n!==arr.lengh; i++) {
+    if (arr.length===n) break;
+    if (n===0) return [];
+    let theBiggest = Math.max(...arr);
+    arr.splice(arr.lastIndexOf(theBiggest), 1);
```
AlyonaPianykh: This function will fail with such arguments` firstNSmallest([1,2,3,4,5], 10)` because length of original array (length = 5) is less than n (n = 10), incorrect condition causes infinite loop. This is not very good practice to use extra variables in the condition in `for` loop (if  you are using `i`, use it fo end condition too, don't use `n`)

By the way, this solution is similar to Eduard's one. Even error int he word `lengh `.

Please rewrite this solution by YOURSELF, probably you can use `while` loop

PR #43, opened 2018-10-22
[Eduard Hrabchak] tasks 41-50

EdGrLab:


tasks/42_first-n-smallest/first-n-smallest.js
```diff
@@ -0,0 +1,9 @@
+export function firstNSmallest(arr, n) {
+  for (let i = 0; n !== arr.lengh; i++) {
```
AlyonaPianykh: This function will fail with such arguments` firstNSmallest([1,2,3,4,5], 10)` because length of original array (length = 5) is less than n (n = 10), incorrect condition causes infinite loop. This is not very good practice to use extra variables in the condition in `for` loop (if  you are using `i`, use it fo end condition too, don't use `n`)

By the way, this solution is similar to Taras' one. Even error int he word `lengh `.

Please rewrite this solution by YOURSELF, probably you can use `while` loop

tasks/45_equal-sides/equal-sides.js
```diff
@@ -0,0 +1,20 @@
...
+  const arrLen = arr.length;
+  let leftsum;
+  let rightsum;
+  for (let i = 0; i < arrLen; i++) {
+    leftsum = 0, rightsum = 0;
+    for (let j = 0; j < i; j++) {
+      leftsum += arr[j];
+    }
+    for (let j = i + 1; j < arrLen; j++) {
+      rightsum += arr[j];
```
AlyonaPianykh: Using several loops makes a mess in code, you can write the separate function to find the sum and apply this function to different parts of array, so that the code will look more beautiful.

Please refactor this.

tasks/50_camel-case/camel-case.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable-next-line */
+String.prototype.camelCase = function() {
```
AlyonaPianykh: Good solution

PR #44, opened 2018-10-22
[TARAS YAKYMIUK]Tasks51 60

taras1313:


tasks/53_diff-array/diff-array.js
```diff
@@ -0,0 +1,9 @@
+/* eslint "require-jsdoc": 0 */
+
+
+export function diffArray(arr1, arr2) {
+  // let res = [];
```
dverbovyi: remove commented code

PR #46, opened 2018-10-22
[A.Zhukevych]Tasks 41-50

Hollister009:


tasks/44_time-simulation/time-simulation.js
```diff
@@ -0,0 +1,25 @@
...
+
+  update(time) {
+    if (this.realtime > time) return 'Error';
+    const newTime = this.time + (time - this.realtime) * this.speed;
+    this.time = newTime;
+    this.realtime = time;
+  }
+}
+
+export {SimTime};
```
dverbovyi: export class SimTime

tasks/48_follow-the-spy/follow-the-spy.js
```diff
@@ -0,0 +1,15 @@
+export const followTheSpy = (directions) => {
+  const firstRoute = directions[0];
+  directions.splice(0, 1);
+  let routes = firstRoute;
+  while (directions.length > 0) {
```
dverbovyi: you don't need while-loop here

Hollister009: My solution doesn't work without it. What should I change also to make it working without while loop ?

dverbovyi: you can check this condition inside for-loop and break the loop when the condition reach success

Hollister009: Trying to solve it without it, but I cannot find a solution yet.

Hollister009: I found the solution!

PR #54, opened 2018-10-24
[NATALIYA_MYKYTYUK] Tasks 41-50

dnistrova:


tasks/42_first-n-smallest/first-n-smallest.js
```diff
@@ -0,0 +1,22 @@
+export function firstNSmallest(array, n) {
+  const result = [];
+  for (let i=0; i<array.length; i++) {
+    let num = array[i];
```
AlyonaPianykh: Nice solution

tasks/43_brackets-indexes/brackets-indexes.js
```diff
@@ -0,0 +1,19 @@
+export function findCloseIndex(str, pos) {
+  if (str[pos] != '(') {
+    return -1;
```
AlyonaPianykh: Very good

tasks/49_cat-and-dog-years/cat-and-dog-years.js
```diff
@@ -0,0 +1,20 @@
+export const humanYearsCatYearsDogYears = function(humanYears) {
+  let catYears;
+  let dogYears;
+  if (humanYears <= 0 || typeof humanYears !== 'number') {
+    return [0, 0, 0];
+  }
```
AlyonaPianykh: Please simplify this function, too much code.
You can use  for example `switch` and  write like this  `catYears = dogYears = 15`
Or you can use array as initial data structure for saving result 
Please think about improving.

PR #57, opened 2018-10-24
[STAS-MOLCHANOV]tasks 41-50

moskdev:


tasks/41_sort-by-bits/sort-by-bits.js
```diff
@@ -0,0 +1,9 @@
+export function sortByBits(arr) {
```
tarasbobak: This solution is very interesting :)

tasks/41_sort-by-bits/sort-by-bits.js
```diff
@@ -0,0 +1,9 @@
...
+  const targetArr = arr.slice();
+  return targetArr.sort((a, b) => {
+    let diff = a.toString(2).replace(
```
tarasbobak: const

tasks/41_sort-by-bits/sort-by-bits.js
```diff
@@ -0,0 +1,9 @@
...
+        /0/g, '') - b.toString(2).replace(/0/g, '');
```
tarasbobak: Do not wrap lines like that.
Better way is to wrap on higher syntaxical level. In this case - . This could be taken into some small utility function

tasks/41_sort-by-bits/sort-by-bits.js
```diff
@@ -0,0 +1,9 @@
...
+    if (diff == 0) return a - b;
```
tarasbobak: use strict comparisons by default: === and !==

tasks/42_first-n-smallest/first-n-smallest.js
```diff
@@ -0,0 +1,12 @@
+export function firstNSmallest(arr, number) {
```
tarasbobak: non-standard, nice

tasks/43_brackets-indexes/brackets-indexes.js
```diff
@@ -0,0 +1,10 @@
+export function findCloseIndex(str, start) {
```
tarasbobak: nice

tasks/44_time-simulation/time-simulation.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+  get() {
+    return this.simulatedTime;
+  }
+  setSpeed(newSpeedRate) {
+    this.speedRate = newSpeedRate;
+  }
+  update(value) {
+    if (value > this.time) {
+      this.simulatedTime = this.simulatedTime + this.speedRate
```
tarasbobak: The operator itself *, + needs to be on previous line according to google styleguide and each next wrapped line needs to have additional 2 pace indentation

tasks/45_equal-sides/equal-sides.js
```diff
@@ -0,0 +1,18 @@
+export function findEqualIndex(arr) {
+  let leftArr = [];
```
tarasbobak: This solution is possible, but your colleagues have got a better one:
Calculate overall sum and them iterate and substract each value. When it will be equal, index of such item will be the index we are looking for.

tasks/47_string-expansion/string-expansion.js
```diff
@@ -0,0 +1,24 @@
+export function stringExpansion(str) {
```
tarasbobak: Loop inside a loop inside a loop :)
You could utilize string.repeat method

tasks/48_follow-the-spy/follow-the-spy.js
```diff
@@ -0,0 +1,14 @@
+export function followTheSpy(arr) {
+  let targetArr = arr.slice();
+  let resultArr = [];
+  const arrLength = arr.length;
+  resultArr.push(arr[0][0], arr[0][1]);
+  while (resultArr.length <= arrLength) {
+    targetArr.map( (value) => {
```
tarasbobak: Mission of a map is to create a new array by mapping old ones values.
So it's not good use for it. You would better use forEach here.

tasks/50_camel-case/camel-case.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable*/
+String.prototype.camelCase = function() {
+  return this.split(' ').map((value, index, array) => {
+    return value = value.slice(0, 1).toUpperCase() + value.slice(1);
```
tarasbobak: good

tarasbobak:
This PR is overly good 👍

PR #58, opened 2018-10-24
[Eduard Hrabchak] tasks 61-70

EdGrLab:


EdGrLab:
Hi @AlexLarusso , thanks for feedback, but I am still wondering about your phrase  "empty line before". Does it mean that I have to skip the next line after } bracket?

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,15 @@
+export function unpackArray(arr) {
```
tarasbobak: Good

PR #59, opened 2018-10-24
[stas-molchanov]tasks 51 to 60

moskdev:


tasks/51_abbreviated-name/abbreviated-name.js
```diff
@@ -0,0 +1,8 @@
+export function abbreviated(str) {
+  return str.split(' ').map((value, index, array) => {
+    if ( index == 0) return value;
+    return value = value.slice(0, 1).toUpperCase() + '.';
+  }
+  ).join(' ');
```
tarasbobak: this should be on previous line, like:
}).join(' ');

tasks/53_diff-array/diff-array.js
```diff
@@ -0,0 +1,6 @@
+export function diffArray(arr1, arr2) {
```
tarasbobak: very good solution

tasks/55_find-object-analogy/find-object-analogy.js
```diff
@@ -0,0 +1,13 @@
+export function whatIsInAName(arr, obj) {
```
tarasbobak: This solution is overcomplicated and also I don't like this: 
boolPush *= value.hasOwnProperty(internalValue) && obj[
          internalValue] == value[internalValue];

Please rewrite it. Here I would go with simple forEach

tasks/56_search-and-replace/search-and-replace.js
```diff
@@ -0,0 +1,12 @@
+export function myReplace(str, before, after) {
+  const arr = str.split('');
+  let insert;
+  const befLength = before.length;
+  if (before.charAt(0).toUpperCase() === before.charAt(0)) {
+    insert = (after.charAt(0).toUpperCase() + after.slice(1)).split('');
```
tarasbobak: so, why do you split string if on line 10 you join it back ?)

tasks/59_sorted-sequence/sorted-sequence.js
```diff
@@ -0,0 +1,9 @@
+export function uniteUnique(...args) {
+  let unique = [];
+  args.map((element, index, array) => {
```
tarasbobak: why map here? map is used to transform array. To iterate over array use for, forEach

tarasbobak:
Please rewrite find-object-analogy.js
Everything else if pretty good. 👍

PR #60, opened 2018-10-24
[Y_Demkiv] Tasks 41 - 50

YarynaD:


tasks/41_sort-by-bits/sort-by-bits.js
```diff
@@ -0,0 +1,23 @@
+export function sortByBits(arr) {
+  let newArr=[];
```
tarasbobak: You are doing it in a pretty complicated way.
You already have arr of numbers and you have to sort them.
I guess you missed my explanation yesterday.

So what you could do is write a function which calculates bitsCount
Then write sort and compare bits of numbers, if they are equal compare numbers themselves.

tasks/42_first-n-smallest/first-n-smallest.js
```diff
@@ -0,0 +1,35 @@
+function getArrayDiff(arr, arrTemp) {
+  let ret = [];
```
tarasbobak: This is also very complicated solution. What you could do is sort from smaller to bigger and then just take needed count.

tasks/43_brackets-indexes/brackets-indexes.js
```diff
@@ -0,0 +1,23 @@
+export function findCloseIndex(str, pos) {
```
tarasbobak: Same here.
indexOf returns first index found.
So you could slice string so that it starts from opening bracket and get index of first closing bracket

tasks/44_time-simulation/time-simulation.js
```diff
@@ -0,0 +1,22 @@
+export class SimTime {
```
tarasbobak: Very good

tasks/46_running-average/running-average.js
```diff
@@ -0,0 +1,9 @@
+export function runningAverage() {
+  let n = 0;
+  let sum = 0;
+  return function(x) {
+    n++;
+    sum += x;
+    return Math.round(sum/n*100)/100;
```
tarasbobak: Will it fail without * 100 / 100 ?

tasks/47_string-expansion/string-expansion.js
```diff
@@ -0,0 +1,14 @@
+export function stringExpansion(str) {
```
tarasbobak: Very good

tasks/48_follow-the-spy/follow-the-spy.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable-next-line */
+Array.prototype.remove = function(from, to) {
```
tarasbobak: This solution is bad, let's work on a dashboard tomorrow morning to discuss.

tasks/49_cat-and-dog-years/cat-and-dog-years.js
```diff
@@ -0,0 +1,20 @@
+export function humanYearsCatYearsDogYears(humanYears) {
```
tarasbobak: Ok, there is a lot of code duplication here, but you will learn it.
This is bad way of writing it catYears = 15 + 9;
Don't you know that 15 + 9 is 24 :)

tasks/49_cat-and-dog-years/cat-and-dog-years.js
```diff
@@ -0,0 +1,20 @@
...
+  if (humanYears === 1) {
+    catYears = 15;
+    dogYears = 15;
+  }
+  if (humanYears === 2) {
+    catYears = 15 + 9;
+    dogYears = 15 + 9;
+  }
+  if (humanYears > 2) {
+    catYears = (24) + (4 * humanYears - 8);
```
tarasbobak: why 24 is in brackets?

tasks/50_camel-case/camel-case.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable-next-line */
+String.prototype.camelCase = function() {
```
tarasbobak: This solution looks great.

tarasbobak:
Some solutions are just awesome, simple and good.
3 are bad. I'm not gonna require you to fix them, but we will discuss them tomorrow morning and draw the way to think on the board.

PR #63, opened 2018-10-24
[Artur Kostiuk] tasks_51-60

art1kos:
tasks_51-60

### DIFF
--- a/tasks/55_find-object-analogy/find-object-analogy.js
+++ b/tasks/55_find-object-analogy/find-object-analogy.js
@@ -1,0 +1,19 @@
+export const whatIsInAName = (arr, source) => {
+  let newArr = [];
+  arr.some((obj) => {
+    const sourceKeys = Object.keys(source);
+    const sourceValues = Object.values(source);
+    let temp = 0;
+    for (let i = 0; i < sourceKeys.length; i++) {
+      if (obj.hasOwnProperty(sourceKeys[i]) &&
+       obj[sourceKeys[i]] === sourceValues[i]) {
+        temp++;
+      }
+    }
+    if (sourceKeys.length === temp) {
+      newArr.push(obj);
+      temp = 0;
+    }
+  });
+  return newArr;
+};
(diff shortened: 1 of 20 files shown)


tasks/60_convert-html-entities/convert-html-entities.js
```diff
@@ -0,0 +1,19 @@
...
+  const obj = {
+    '&': '&amp;',
+    '<': '&lt;',
+    '>': '&gt;',
+    '"': '&quot;',
+    '\'': '&apos;',
+  };
+  for (let i = 0; i < strArrLength; i++) {
+    for (let prop in obj) {
+      if (strArr[i] === prop) {
```
prezlyataa: You need here additional condition. A `for ... in` loop iterates own object's properties and properties from `Object.prototype`. You should iterate only own properties, use method hasOwnProperty to verify it. For example: `if(obj.hasOwnProperty(prop) && strArr[i] === prop)`

tasks/55_find-object-analogy/find-object-analogy.js
```diff
@@ -0,0 +1,19 @@
+export const whatIsInAName = (arr, source) => {
+  let newArr = [];
+  arr.some((obj) => {
+    const sourceKeys = Object.keys(source);
+    const sourceValues = Object.values(source);
+    let temp = 0;
```
prezlyataa: Please declare variables const and let on the beginning of the function, where you declared newArr.

tasks/56_search-and-replace/search-and-replace.js
```diff
@@ -0,0 +1,15 @@
+export const myReplace = (str, str1, str2) => {
+  const strArr = str.split(' ');
+  if (strArr.indexOf(str1)) {
+    const firstLetter = str1.charAt(0);
+    const check = str1.startsWith(firstLetter.toUpperCase());
+    const firstLetterToUpper = str2.charAt(0).toUpperCase() + str2.slice(1);
```
prezlyataa: In this case also, please declare all const on the beginning of the function.

PR #67, opened 2018-10-25
[Natalia Falinska] Tasks 41-50

NFalinska:


tasks/50_camel-case/camel-case.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable*/
+String.prototype.camelCase = function() {
+  let newArr = this.split(' ');
+  for (let i = 0; i < newArr.length; i++) {
+    newArr[i] = newArr[i].charAt(0).toUpperCase() + newArr[i].slice(1);
```
AlexLarusso: use template string

tasks/50_camel-case/camel-case.js
```diff
@@ -0,0 +1,8 @@
...
+  }
+  return newArr.join('');
```
AlexLarusso: empty line before

tarasbobak:
This needs to be recreated to proper branch!

PR #68, opened 2018-10-25
[Eduard Hrabchak] Tasks: 71-80

EdGrLab:


tasks/72_carrying/carrying.js
```diff
@@ -0,0 +1,11 @@
+export function addTogether(...args) {
+  const arr = [...args];
+  if (arr.some((el) => typeof el === 'string')) return undefined;
+  else if (arr.length > 1) {
```
Nazar-bnm: you shouldn't use `else` after `return`

tasks/73_check-for-palindromes/check-for-palindromes.js
```diff
@@ -0,0 +1,6 @@
+export function palindrome(str) {
+  let regex = /[\W_]/g;
```
Nazar-bnm: use const

tasks/73_check-for-palindromes/check-for-palindromes.js
```diff
@@ -0,0 +1,6 @@
...
+  let lowRegStr = str.toLowerCase().replace(regex, '');
```
Nazar-bnm: use const (you never change it)

tasks/79_calculate_primoral_of_a_number/calculate_primoral_of_a_number.js
```diff
@@ -0,0 +1,19 @@
+export function numPrimorial(num) {
+  let newArr = [];
```
Nazar-bnm: const

PR #69, opened 2018-10-25
[Y_Demkiv] Tasks 51-60

YarynaD:


tasks/53_diff-array/diff-array.js
```diff
@@ -0,0 +1,9 @@
+export function diffArray(arr1, arr2) {
+  const newArr = arr1.concat(arr2);
+  function check(item) {
+    if (arr1.indexOf(item) === -1 || arr2.indexOf(item) === -1) {
+      return item;
+    }
+  }
+  return newArr.filter(check);
```
AlyonaPianykh: You can simplify this function like this:
`
  const newArr = arr1.concat(arr2);
  return newArr.filter(item => (arr1.indexOf(item) === -1 || arr2.indexOf(item) === -1));
`

tasks/55_find-object-analogy/find-object-analogy.js
```diff
@@ -0,0 +1,11 @@
+export function whatIsInAName(collection, source) {
+  let array = [];
+  let keys = Object.keys(source);
+  array = collection.filter(function(obj) {
+    return keys.every(function(key) {
```
AlyonaPianykh: Very good

tasks/56_search-and-replace/search-and-replace.js
```diff
@@ -0,0 +1,9 @@
+export function myReplace(str, before, after) {
+  return str.replace(before, function(before) {
```
AlyonaPianykh: It's better to use another word and not "before" as a function argument (second argument of `str.replace`)because the outer function already has an argument with such name. It makes code very hard to understand

tasks/59_sorted-sequence/sorted-sequence.js
```diff
@@ -0,0 +1,6 @@
+export function uniteUnique() {
+/* eslint-disable-next-line */
+  return [].concat.apply([], arguments).filter(function(elem, index, self) {
+    return self.indexOf(elem) === index;
+  });
```
AlyonaPianykh: Awesome solution

tasks/51_abbreviated-name/abbreviated-name.js
```diff
@@ -0,0 +1,12 @@
...
+  let arrNew = [];
+  for (let i = 0; i < arr.length; i++) {
+    if (i === 0) {
+      arrNew.push(arr[0]);
+    } else {
+      arrNew.push(arr[i][0] + '.');
+    }
+  }
+  return arrNew.join(' ');
+}
```
AlyonaPianykh: You don't need this condition 
`
if (i === 0) {
      arrNew.push(arr[0]);
    }
` because you can set `let arrNew = [arr[0]];` and then start loop from 1
Please rewrite this function a little

tasks/51_abbreviated-name/abbreviated-name.js
```diff
@@ -1,12 +1,8 @@
 export function abbreviated(str) {
   const arr = str.split(' ');
-  let arrNew = [];
-  for (let i = 0; i < arr.length; i++) {
-    if (i === 0) {
-      arrNew.push(arr[0]);
-    } else {
-      arrNew.push(arr[i][0] + '.');
-    }
+  let arrNew = [arr[0]];
```
tarasbobak: you should use const when there is no reassignment

PR #71, opened 2018-10-25
[Artur Kostiuk]tasks_61-70

art1kos:
tasks_61-70


tasks/67_find-elements/find-elements.js
```diff
@@ -0,0 +1,8 @@
+export const findElement = (arr, cond) => {
+  const arrLength = arr.length;
+  for (let i = 0; i < arrLength; i++) {
```
Nazar-bnm: Set some value to variable when it use multiple times, so you can just use `arr.length` in loop.

PR #73, opened 2018-10-26
[Pavlo T.]Task 21-30

FabianMoretti:


tasks/10_remove-whitespaces/remove-whitespaces.js
```diff
@@ -1,6 +0,0 @@
-/* eslint "require-jsdoc": 0 */
```
tarasbobak: Why is this deleted?

PR #76, opened 2018-10-26
[Y_Demkiv] Tasks 61-70

YarynaD:


tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,25 @@
...
+  for (let i = 1; i <= num; i++) {
+    arrNew.push(i);
+  }
+  let arrPrime = [];
+  for (let i = 0; i <= arrNew.length; i++) {
+    if (isPrime(arrNew[i])) {
+      arrPrime.push(arrNew[i]);
+    }
+  }
+  let sum = arrPrime.reduce(function(accumulator, currentValue) {
```
tarasbobak: No need for sum variable, you can simply return arrPrime.reduce(...

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,25 @@
...
+  for (let i = 2; i < value; i++) {
+    if (value % i === 0) {
+      return false;
+    }
+  }
+  return value > 1;
+}
+
+export function sumPrimes(num) {
+  let arrNew = [];
```
tarasbobak: const

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,25 @@
...
+  for (let i = 1; i <= num; i++) {
+    arrNew.push(i);
+  }
+  let arrPrime = [];
```
tarasbobak: const

tasks/67_find-elements/find-elements.js
```diff
@@ -0,0 +1,7 @@
+export function findElement(arr, func) {
```
tarasbobak: Well, this solution is wrong.
You are hardcoding the logic. But any function could be used to filter out the results.
For instance such function could be used findElement([1, 3, 5, 8, 9, 10], function(num) { return num > 5; }

tasks/67_find-elements/find-elements.js
```diff
@@ -0,0 +1,7 @@
...
+  for (let i = 0; i < arr.length; i++) {
+    if (arr[i] % 2 === 0) {
```
tarasbobak: You should apply function to each element in array and if it results to truthy value it should be returned.

tasks/68_drop-elements/drop-elements.js
```diff
@@ -0,0 +1,10 @@
+export function dropElements(arr, func) {
```
tarasbobak: smart solution

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,12 @@
+export function unpackArray(arr) {
```
tarasbobak: This will work with arrays only one level deep and will fails if you will test with more nested arrays.

For it to work you should have implemented a recursive solution

tasks/70_binary-decode/binary-decode.js
```diff
@@ -0,0 +1,15 @@
...
+  if (!str) {
+    str = '';
+  }
+  let split = str.split(' ');
+  let newArray = [];
+  let result = [];
+
+  for (let i = 0; i < split.length; i++) {
+    newArray.push(parseInt(split[i], 2));
+    result.push(String.fromCharCode(newArray[i]));
```
tarasbobak: here I don't fully understand why you need 2 separate arrays newArray and result.

tarasbobak:
Keep going Yaryna. I don't want to distract you from solving further tasks.
Overly good implementation, but 2 tasks were not good (67 and 69).

PR #80, opened 2018-10-26
[stas-molchanov]tasks 61-70

moskdev:


tasks/63_insert-string/insert-string.js
```diff
@@ -0,0 +1,13 @@
+export function insert(str, ...args) {
+  const argsLength = args.length;
+  let targetArr = str.split('');
+  switch (argsLength) {
```
tarasbobak: Nice usage of switch case here

tasks/63_insert-string/insert-string.js
```diff
@@ -0,0 +1,13 @@
+export function insert(str, ...args) {
+  const argsLength = args.length;
+  let targetArr = str.split('');
```
tarasbobak: const

tasks/63_insert-string/insert-string.js
```diff
@@ -0,0 +1,13 @@
...
+  switch (argsLength) {
+    case 1:
+      return args[0] + str;
+    case 2:
+      targetArr.splice(args[1], 0, args[0].split(''));
+      return targetArr.toString().replace(/,/g, '');
```
tarasbobak: why not targetArr.join('') ?

tasks/64_sum-odd-fibonacci/sum-odd-fibonacci.js
```diff
@@ -0,0 +1,9 @@
+export function sumFibs(number) {
```
tarasbobak: Smart 👍

tasks/65_smaller-common-multiple/smaller-common-multiple.js
```diff
@@ -0,0 +1,17 @@
+export function smallestCommons(args) {
```
tarasbobak: This is very non-standard solution. Generally it is very interesting, but this part is pretty difficult to understand:
let flag =false;
  while (!flag) {
    commonMult += max;
    flag = targetArr.reduce((acum, value) =>{
      return acum * (!(commonMult % value));
    }, max);
  }

In real life it is generally better to write code, that is easy to understand. Because you write code once, but it will be read by other developers multiple times

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,17 @@
+export function sumPrimes(n) {
+  const primeArr = [2];
+  for ( let i = 3; i <= n; i+=2 ) {
+    if ( isPrime(i) ) {
+      primeArr.push(i);
+    }
+  }
+  function isPrime(num) {
+    for ( let i = 2; i < num; i++ ) {
```
tarasbobak: for (let i = 2; i < num; i++) {

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,17 @@
+export function sumPrimes(n) {
+  const primeArr = [2];
+  for ( let i = 3; i <= n; i+=2 ) {
+    if ( isPrime(i) ) {
```
tarasbobak: if (isPrime(i)) {

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,17 @@
...
+      primeArr.push(i);
+    }
+  }
+  function isPrime(num) {
+    for ( let i = 2; i < num; i++ ) {
+      if ( num % i === 0 ) {
```
tarasbobak: if (num % i === 0) {

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,18 @@
...
+  const targetArr = args.slice();
+  const resultArr = [];
+  let pushValue;
+  targetArr.map((element) => {
+    pushValue = Array.isArray(element) ? getInnerArray(element) : element;
+    resultArr.push(pushValue);
+  });
+  function getInnerArray(arr) {
+    return arr.reduce((accum, value) => {
+      if (!Array.isArray(value)) return accum.concat(value);
```
tarasbobak: if () {
} else {
}

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,18 @@
+export function unpackArray(...args) {
+  const targetArr = args.slice();
+  const resultArr = [];
+  let pushValue;
+  targetArr.map((element) => {
```
tarasbobak: I've left you same comment in one of previous PR's
It is wrong usage of map!
You should use it only when you need array transformation into another value. Read MDN for more info

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,18 @@
+export function unpackArray(...args) {
```
tarasbobak: Solution is generally fine

tasks/70_binary-decode/binary-decode.js
```diff
@@ -0,0 +1,9 @@
+export function binaryDecode(str) {
+  if (!str) return '';
+  const codedArr = str.split(' ');
+  const decodedArr = [];
+  codedArr.map((value, index) =>{
```
tarasbobak: Again, not map, but forEach

tarasbobak:
@moskdev I like you solutions. There are mistakes which I wouldn't like to see any more.
Pay attention to comments please.

PR #84, opened 2018-10-27
[Natalia Falinska] Tasks 61_70

NFalinska:


tasks/61_spinal-case/spinal-case.js
```diff
@@ -0,0 +1,4 @@
+export function spinalCase(str) {
+  return str.replace(/(?!^)([A-Z])/g, ' $1')
```
tarasbobak: I saw identical solution.

tasks/62_swap-cases/swap-cases.js
```diff
@@ -0,0 +1,12 @@
+export function swapCases(word) {
```
tarasbobak: Pretty good 👍

tasks/63_insert-string/insert-string.js
```diff
@@ -0,0 +1,5 @@
+export function insert(str, word, item) {
+  if (!word) return str;
+  if (!item) item = 0;
```
tarasbobak: Please read about default params https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,16 @@
...
+    }
+    return num !== 1;
+  }
+  for (let j = 0; j<= num; j++) {
+    if (isPrime(j)) {
+      primeArr.push(j);
+    }
+  }
+  const allSum = primeArr.reduce((a, b)=> a + b);
+  return allSum;
```
tarasbobak: No need for allSum variable.
Just return primeArr.reduce((a, b)=> a + b);

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,12 @@
+export function unpackArray(arr) {
```
tarasbobak: This solution is wrong. It will fail when you will have nested array more than 1 level deep.
Proper solution needs to use recursion.

tasks/70_binary-decode/binary-decode.js
```diff
@@ -0,0 +1,4 @@
+export function binaryDecode(str) {
```
tarasbobak: Smart

tasks/70_binary-decode/binary-decode.js
```diff
@@ -0,0 +1,4 @@
...
+  if (!str) return '';
+  return str.replace(/\d+./g, (x)=>String.fromCharCode('0b'+x));
```
tarasbobak: return str.replace(/\d+./g, x => String.fromCharCode('0b' + x));

tarasbobak:
Please read the comments carefully.
Good job!

PR #89, opened 2018-10-29
[Pavlo Turchyn] tasks:61-70

FabianMoretti:


tasks/61_spinal-case/spinal-case.js
```diff
@@ -0,0 +1,5 @@
+/* eslint "require-jsdoc": 0 */
+export const spinalCase = (str) => {
+  return str.split(/\s|_|(?=[A-Z])/).map((word) => word[0].toLowerCase() +
+   word.substring(1) + '-').join('').slice(0, -1);
```
tarasbobak: use proper indentation
```
return str.split(/\s|_|(?=[A-Z])/).map((word) => word[0].toLowerCase() +
  word.substring(1) + '-').join('').slice(0, -1);
```

tasks/62_swap-cases/swap-cases.js
```diff
@@ -0,0 +1,14 @@
+/* eslint "require-jsdoc": 0 */
+export const swapCases = (str) => {
+  let res = '';
+  const strLength = str.length;
+  for (let i = 0; i < strLength; i++) {
+    let char = str.charAt(i);
```
tarasbobak: const

tasks/62_swap-cases/swap-cases.js
```diff
@@ -0,0 +1,14 @@
...
+    if (char.toUpperCase() === char) {
+      res += char.toLowerCase();
+    } else {
+      res += char.toUpperCase();
+    }
+  }
+  return res + '';
```
tarasbobak: it is already a string, just return res;

tasks/63_insert-string/insert-string.js
```diff
@@ -0,0 +1,7 @@
+/* eslint "require-jsdoc": 0 */
+export const insert = (str, text, i = 0) => {
+  if (!text) return str;
+  let strArr = str.split('');
```
tarasbobak: const

tasks/64_sum-odd-fibonacci/sum-odd-fibonacci.js
```diff
@@ -0,0 +1,27 @@
...
+  let res = 0;
+
+  if (f0 <= num) {
+    res += f0;
+  }
+
+  if (f1 <= num) {
+    res += f1;
+  }
+  while (res) {
```
tarasbobak: Why not while (f1 <= num)

tasks/65_smaller-common-multiple/smaller-common-multiple.js
```diff
@@ -0,0 +1,29 @@
+/* eslint "require-jsdoc": 0 */
+export const smallestCommons = (nums) => {
+  let numsArr = [];
+  if (nums[0] < nums[1]) {
```
tarasbobak: This is a total duplication of code, which is very bad

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,22 @@
+/* eslint "require-jsdoc": 0 */
+export const sumPrimes = (num) => {
+  const isPrime = (value) => {
+    const _length = value / 2 + 1;
```
tarasbobak: This task is copy pasted from someone else or from internet

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,22 @@
...
+    for (let i = 2; i <= _length; i++) {
+      if (value % i == 0 && i != value) return false;
```
tarasbobak: Always use strict comparisons

tasks/66_sum-primes/sum-primes.js
```diff
@@ -0,0 +1,22 @@
...
+    }
+    return true;
+  };
+  let resArr = [];
```
tarasbobak: const

tasks/67_find-elements/find-elements.js
```diff
@@ -0,0 +1,8 @@
+/* eslint "require-jsdoc": 0 */
+export const findElement = (arr, func) => {
+  let res = arr.filter((current) => func(current));
+  if (res === arr) {
```
tarasbobak: WAT???)
Two array will never be equal if they don't share the same reference
Run in console
[1, 2, 3] === [1, 2, 3]

tasks/67_find-elements/find-elements.js
```diff
@@ -0,0 +1,8 @@
+/* eslint "require-jsdoc": 0 */
+export const findElement = (arr, func) => {
+  let res = arr.filter((current) => func(current));
```
tarasbobak: you should have used forEach and return result when any item will fit the function condition.
If nothing would fit undefined would be returned from a function by default;

tasks/68_drop-elements/drop-elements.js
```diff
@@ -0,0 +1,12 @@
+/* eslint "require-jsdoc": 0 */
+export const dropElements = (arr, func) => {
+  let end;
+  const arrLength = arr.length;
+  for (let i = 0; i <= arrLength; i++) {
+    if (func(arr[i]) === true) {
+      end = i;
```
tarasbobak: Why do you need end at all???

tasks/68_drop-elements/drop-elements.js
```diff
@@ -0,0 +1,12 @@
+/* eslint "require-jsdoc": 0 */
+export const dropElements = (arr, func) => {
+  let end;
+  const arrLength = arr.length;
+  for (let i = 0; i <= arrLength; i++) {
+    if (func(arr[i]) === true) {
```
tarasbobak: if (func(arr[i])) {

tarasbobak: Any value inside if conditional statement will be by default checked if it's truthy

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,13 @@
+/* eslint "require-jsdoc": 0 */
+export const unpackArray = (arr) => {
+  let res = [];
```
tarasbobak: const

tasks/69_unpack-elements/unpack-elements.js
```diff
@@ -0,0 +1,13 @@
+/* eslint "require-jsdoc": 0 */
+export const unpackArray = (arr) => {
```
tarasbobak: Solution is interesting 👍

tasks/70_binary-decode/binary-decode.js
```diff
@@ -0,0 +1,19 @@
+/* eslint "require-jsdoc": 0 */
+export const binaryDecode = (str) => {
+  let convertBinaryToString;
+  let deleteSpaces;
+  if (!str) {
+    str = '';
+  }
+  convertBinaryToString = str.replace(/[01]{8}/g, (element) => {
```
tarasbobak: This is sooo complicated... Why not simply chain replaces?

tarasbobak:
Generally good work Pavlo, but there is a lot to improve.
You sometimes create not needed variables, do code duplication, not properly use const and let.

PR #98, opened 2018-10-30
[NATALIYA_MYKYTYUK] Tasks 81-90

dnistrova:


tarasbobak:
Needs to be recreated -> should not be merged into master

PR #147, opened 2018-11-20
[Artur Kostiuk]css-practice-task3

art1kos:
https://art1kos.github.io/ui-internship-lwo2/css-practice/task_3/index.html#


tarasbobak:
Fix font size in heading section

PR #154, opened 2018-11-21
[Y_Demkiv] css practice task 3

YarynaD:
https://yarynad.github.io/ui-internship-lwo2/css-practice/task_3/index.html


tarasbobak:
Generally pretty good.

I see some flows on ipad and generally
screen shot 2018-11-22 at 10 32 55 am
screen shot 2018-11-22 at 10 33 09 am

But mobile is good.

Regarding hover you should have implemented it yourself as a dynamic effect.

PR #159, opened 2018-11-23
[TARAS-YAKYMIUK]first-version-validation

taras1313:
https://taras1313.github.io/ui-internship-lwo2/css-practice/task2-2/index.html


tarasbobak:
Please fix discussed comments.

css-practice/task2-2/index.js
```diff
@@ -0,0 +1,127 @@
+/* eslint-disable */
+(function() {
+
+  const menu = document.querySelector('.menu-btn');
+  const hiddenMenu = document.querySelector('.burger-menu');
+  menu.onclick = () => {
+    hiddenMenu.classList.contains('d-none') ? hiddenMenu.classList.remove('d-none') : hiddenMenu.classList.add('d-none');
```
AlyonaPianykh: You can use ```toggle```

css-practice/task2-2/index.js
```diff
@@ -0,0 +1,124 @@
...
+        tabsP.classList.toggle('opacity-0');
+      }, 200)
+    }
+  }
+
+
+
+  const REGEXP = {
+    user_name: {
+      reg: /^[a-zA-Z ]{2,45}$/,
```
AlyonaPianykh: You should remove validation from name and leave only non empty, cause there are cases when parents called there child "R2D2" or like here https://ru.wikipedia.org/wiki/%D0%91%D0%9E%D0%A7_%D1%80%D0%92%D0%A4_260602

PR #169, opened 2018-11-27
[Sophia] Second template not axit (video, disabled button)

SofiMat:


SofiMat:
https://sofimat.github.io/ui-internship-lwo2/css-practice/task2/

css-practice/task2/script.js
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable */
```
tarasbobak: why eslint-disable???

css-practice/task2/script.js
```diff
@@ -0,0 +1,89 @@
...
+  const ERROR_CLASSNAME = 'error';
+
+  handleFormValidation(['contact-form']);
+  FORMS_ON_PAGE.forEach((el) => {
+    el.addEventListener('submit', submit);
+  });
+
+  function handleFormValidation(formName) {
+    const form = document.forms[formName];
+    form.addEventListener('keyup', (e) => validateInput(e.target), true);
```
tarasbobak: Why do you need 2 separate addEventListener's???

css-practice/task2/script.js
```diff
@@ -0,0 +1,89 @@
...
+  function resetForm(input) {
+    if (input.value === '') {
+      input.classList.remove('valid');
+      input.classList.remove('error');
+    }
+  }
+
+  function submit(e) {
+    const inputs = [...this.elements];
+    if (inputs.some((el) => (el.value === '' && el['type'] !== 'submit') || isValidValue(el.value, PATTERNS[el.name]) === false)) {
```
tarasbobak: Move everything inside this if to a variable and give it proper name

css-practice/task2/script.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+      });
+    }
+  }
+  
+  const buttonSub = document.getElementById('submit-button');
+
+  function disable() {
+    const inputs = [...document.forms['contact-form'].querySelectorAll('input')];
+    if (inputs.some(el => el.classList.contains('error') || el.value === '')) {
```
tarasbobak: Move to variable

tarasbobak: I mean const in both cases

tarasbobak: Readibility is why we need to do this.

css-practice/task2/script.js
```diff
@@ -0,0 +1,89 @@
...
+      buttonSub.classList.add('disabled');
+    } else {
+      buttonSub.classList.remove('disabled');
+    }
+  }
+  disable();
+})();
+
+(() => {
```
tarasbobak: It looks you misunderstood a bit what I was explaining. You don't need to uses IIFE all the time to encapsulate.
Better in your case to create a function names initVideo and then invoke it. Same goes for initFormValidation

tarasbobak:
Please fix comments

css-practice/task2/script.js
SofiMat: in that case my "document" is undefined

css-practice/task2/script.js
SofiMat: removed one

css-practice/task2/script.js
SofiMat: updated

css-practice/task2/script.js
SofiMat: done

css-practice/task2/script.js
SofiMat: i change it, hope it was what you meant

css-practice/task2/script.js
tarasbobak: you should declare globals with eslint comments /* global var1, var2 */

css-practice/task2/script.js
```diff
@@ -49,21 +49,34 @@
...
   const buttonSub = document.getElementById('submit-button');
 
   function disable() {
     const inputs = [...document.forms['contact-form'].querySelectorAll('input')];
-    if (inputs.some(el => el.classList.contains('error') || el.value === '')) {
+
+    const invalid = function(el) {
+      return el.classList.contains('error') || el.value === '';
+    }
+    if (inputs.some((el) => invalid(el))) {
```
tarasbobak: this can be also stored in some variable with defined and meaningful name

css-practice/task2/script.js
```diff
@@ -72,18 +85,17 @@
   disable();
 })();
 
-(() => {
-const videoState = document.querySelector('.player');
-const pauseVideo = document.querySelector('.fa-pause-circle');
-const playVideo = document.querySelector('.fa-play-circle');
-
-videoState.addEventListener('click', pauseVideoFunc);
+(function initVideo() {
```
tarasbobak: this is not exactly what I meant as IIFE is kinda old-school approach.
Anyways for now it can be like that.

css-practice/task2/script.js
```diff
@@ -72,18 +85,17 @@
...
-  pauseVideo.classList.toggle('none');
-  playVideo.classList.toggle('none');
+  videoState.addEventListener('click', pauseVideoFunc);
 
-}
-})()
+  function pauseVideoFunc() {
+    const video = document.querySelector('video');
+    video.paused ? video.play() : video.pause();
+    pauseVideo.classList.toggle('none');
```
tarasbobak: every string that is used more than once has to be stored in const

css-practice/task2/script.js
```diff
@@ -72,18 +85,17 @@
...
-  const video = document.querySelector('video');
-  video.paused ? video.play() : video.pause();
-  pauseVideo.classList.toggle('none');
-  playVideo.classList.toggle('none');
+  videoState.addEventListener('click', pauseVideoFunc);
 
-}
-})()
+  function pauseVideoFunc() {
+    const video = document.querySelector('video');
```
tarasbobak: After variable declaration it is nice to have an empty line to give a bit more "space" for code

tarasbobak:
Sofia it is already pretty good code, but please fix new comments and I will merge it

css-practice/task2/script.js
SofiMat: updated

css-practice/task2/script.js
SofiMat: I changed it, if in this case it's still not what you meant, than please explain a good practice :)

css-practice/task2/script.js
SofiMat: I couldn't see any which is used more that once, which one do you mean?

css-practice/task2/script.js
SofiMat: done

tarasbobak:
Looks much better now.
Last recommendations:
1) Don't wrap lines if they do not exceed maximum line length
2) Start naming you commits properly
screen shot 2018-12-04 at 12 53 00 pm

PR #170, opened 2018-11-27
[Eduard Hrabchak] Interactive task 2

EdGrLab:
https://edgrlab.github.io/ui-internship-lwo2/css-practice/task2/


tarasbobak:
This seems like has to be fixed
screen shot 2018-12-02 at 8 55 50 pm

tarasbobak:
Firm implementation is wrong.
Button has to be disabled until form is valid.
By default it is disabled as form is not valid

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable */
```
tarasbobak: Remove eslint disable and specify global instead

To specify globals using a comment inside of your JavaScript file, use the following format:

/* global var1, var2 */

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+const playVideo = document.querySelector(".video-player");
+const contineVideo = document.querySelector(".fa-play-circle");
+const pauseVideo = document.querySelector(".fa-pause-circle");
+
+playVideo.addEventListener("click", pauseVid, false);
```
tarasbobak: false is default so no need to pass it

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+function pauseVid() {
+  const video = document.querySelector("video");
+  video.paused ? video.play() : video.pause();
+  pauseVideo.classList.toggle("hide-el");
```
tarasbobak: every string used more than once has to be stored in const

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable */
+
+const playVideo = document.querySelector(".video-player");
+const contineVideo = document.querySelector(".fa-play-circle");
+const pauseVideo = document.querySelector(".fa-pause-circle");
+
+playVideo.addEventListener("click", pauseVid, false);
+
+function pauseVid() {
+  const video = document.querySelector("video");
```
tarasbobak: in javascript always use single quotes!

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
...
+  pass: /(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z])(?=.*[!@#$%^&+-]).{6,20}/
+};
+
+function validateForm(id, e) {
+  const inputArr = Array.from(document.getElementById(id).children);
+  for (let i = 0; i < inputArr.length; i++) {
+    if (inputArr[i].classList.contains("invalid")) {
+      btnDisable.classList.add("disable-submit");
+      e.preventDefault();
+      return false;
```
tarasbobak: return false in this case will call event.preventDefault() for you so no need to duplicates this

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+const btnDisable = document.getElementById("submit-btn");
+
+const patterns = {
+  name: /^[a-z]{3,}/,
+  email: /^[^@]+@[^@]+.[^@]{2,}$/,
+  pass: /(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z])(?=.*[!@#$%^&+-]).{6,20}/
+};
+
+function validateForm(id, e) {
```
tarasbobak: if this is form id call it formId

css-practice/task2/index.js
```diff
@@ -0,0 +1,53 @@
...
+      e.preventDefault();
+      return false;
+    }
+  }
+}
+
+function validate(ev) {
+  const pattern = patterns[ev.target.name];
+  if (!ev.target.value) {
+    ev.target.classList.remove("valid");
```
tarasbobak: you duplicate these 2 lines of code 2 times.
I would create 2 short functions: 1- markInvalid, 2- markValid
And simplify logic here.

tarasbobak:
Fix comments, resolve conflicts, implement proper form logic.

css-practice/task2/index.js
```diff
@@ -1,53 +1,72 @@
...
-      btnDisable.classList.add("disable-submit");
-      e.preventDefault();
+    if (inputArr[i].classList.contains('invalid')) {
+      btnDisable.classList.add('disable-submit');
       return false;
     }
   }
 }
 
+function markInvalid(ev) {
```
tarasbobak: This is not what I meant markInvalid would do 
ev.target.classList.remove('valid');
and
ev.target.classList.add('invalid'); in the same function

markValid would do the opposite

tarasbobak:
Conflicts are not resolved.
Form logic is still wrong - button is not disabled. Perhaps you haven't deployed new version to gh-pages.
This is not fixed:
screen shot 2018-12-02 at 8 55 50 pm

tarasbobak:
Please fix comments carefully.

tarasbobak:
Unfortunately you didn't fix form validation. But I'm merging since we need to move on.

PR #171, opened 2018-11-28
[NATALIYA_MYKYTYUK] Video and Form

dnistrova:
https://dnistrova.github.io/ui-internship-lwo2/css-practice/task_4/index.html


css-practice/task_4/script.js
```diff
@@ -0,0 +1,41 @@
+const video = document.getElementsByClassName('main-video')[0];
+const videoBtn = document.getElementsByClassName('play-mode')[0];
+const formInput = document.querySelectorAll('input');
```
tarasbobak: Are you looking for all inputs on page or inputs inside specific form?

css-practice/task_4/script.js
```diff
@@ -0,0 +1,41 @@
...
+const inputs = Array.from(formInput);
+const form = document.forms[0];
+const tryNowBtn = document.getElementById('form-button');
+
+function pauseVideo() {
+  if (!video.paused) {
+    video.pause();
+    videoBtn.classList.remove('play');
```
tarasbobak: whenever you use any string more than once it has to be moved to const

css-practice/task_4/script.js
```diff
@@ -0,0 +1,41 @@
+const video = document.getElementsByClassName('main-video')[0];
+const videoBtn = document.getElementsByClassName('play-mode')[0];
+const formInput = document.querySelectorAll('input');
+const inputs = Array.from(formInput);
+const form = document.forms[0];
+const tryNowBtn = document.getElementById('form-button');
+
+function pauseVideo() {
```
tarasbobak: function name is pauseVideo but it pauses and continues, name is not representing what this function does.
You should strife to write function with single responsibility.
So ideally function to handle pause/play button click.
Inside you will have if statements which handles 2 other function calls, one to play video and next to stop. Something like that.

tarasbobak: Ideally this has to be implemented as incapsulated module with init function exposed as API.
But this is not required.
If you will have time and will this will be good exercise for you

css-practice/task_4/script.js
```diff
@@ -0,0 +1,41 @@
...
+    video.play();
+    videoBtn.classList.add('play');
+    videoBtn.classList.remove('pause');
+  }
+}
+
+videoBtn.addEventListener('click', pauseVideo);
+
+function validatePattern(event) {
+  const element = event.target;
```
tarasbobak: it is nice to have empty line after variable declarations to give code some space

css-practice/task_4/script.js
```diff
@@ -0,0 +1,41 @@
...
+    element.classList.add('invalid');
+  } else {
+    element.classList.remove('invalid');
+    element.classList.add('valid');
+  }
+}
+
+form.addEventListener('keyup', validatePattern, true);
+function notDisabled() {
+  if (inputs.every((el) => el.classList.contains('valid'))) {
```
tarasbobak: move this to const and use it in if statement instead.

css-practice/task_4/script.js
```diff
@@ -0,0 +1,41 @@
...
+    tryNowBtn.classList.remove('disabled');
+  }
+}
+// tryNowBtn.addEventListener('click', notDisabled);
```
tarasbobak: remove comment

tarasbobak:
Resolve conflict, fix comments.

tarasbobak:
It is pretty good now.
Last comment - it is not a good practice to notify user he typed invalid email when he only starts typing in an input field and this input has not been touched before.

2 ways of how it is usually done:
1) we wait until user tries to submit a form which is not your case as we want button disabled
2) we validate for the first time on blur. And after that we can notify user in real time

PR #177, opened 2018-11-29
[A.Zhukevych]Interactive-task 2

Hollister009:
https://hollister009.github.io/ui-internship-lwo2/css-practice/cooperative-task/


css-practice/cooperative-task/js/script.js
```diff
@@ -0,0 +1,55 @@
...
+  // select inputs from form
+  const inputsArr = [...contactForm.elements].filter((el) => {
+    return el.type === 'text' || el.type === 'email' || el.type === 'password';
+  });
+
+  const checkUserInput = (input) => {
+    const value = input.value;
+    const pattern = new RegExp(input.getAttribute('data-pattern'));
+    const isValid = pattern.test(value);
+    if (isValid) {
```
prezlyataa: You should add additional condition because when you fill your fields and after that clean them, colors are still active, it must be just default form without highlights.

css-practice/cooperative-task/js/script.js
```diff
@@ -0,0 +1,55 @@
...
+    } else {
+      input.classList.remove('valid');
+      input.classList.add('invalid');
+    }
+  };
+
+  const submitSwitch = (e) => {
+    const input = e.target;
+    checkUserInput(input);
+    const subState = inputsArr.every((el) => {
```
prezlyataa: All constants should be declared on the top.

css-practice/cooperative-task/js/script.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+/* global document alert */
+(function() {
+  // Background video playback
+  const playBtn = document.querySelector('.btn-video');
+  const bgVideo = document.querySelector('video');
+
+  const playState = (evt) => {
+    evt.preventDefault();
+    if (playBtn.classList.contains('play')) {
```
prezlyataa: When video is stopped your button contain state play, but in this case it must be vice versa.

css-practice/cooperative-task/js/script.js
Hollister009: In that case _checkUserInput(input)_ will not run properly, if I put it after _const subState_

css-practice/cooperative-task/js/script.js
Hollister009: Changed

PR #178, opened 2018-11-29
[Y_Demkiv] Task 2 video run and form disable

YarynaD:
https://yarynad.github.io/ui-internship-lwo2/css-practice/task_2/index.html


dverbovyi:
It seems to me the button is not aligned properly
screen shot 2018-12-03 at 15 53 16

css-practice/task_2/form.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable */
```
tarasbobak: Well, generally I don't like this implementation. It is for 3 out of 5.
1) you attach event listeners to every input instead of 1 for a form
2) you style valid/invalid from javascript, but I prefer to have classes with styles to add and remove
3) newDisableFunc - why is it declared without var, const? If you want function declaration it is written like function newDisableFunc. And why new, where is old in file?

css-practice/task_2/form.js
```diff
@@ -0,0 +1,36 @@
...
+
+const myForm = document.querySelectorAll('form,input');
+btn.disabled = true;
+
+newDisableFunc = function() {
+  if (test_rules[this.name].test(this.value)) {
+    this.style.outline = '2px solid green';
+    test_passed[this.name] = true;
+    btn.disabled = false;
+    btn.classList.remove('disabled');
```
tarasbobak: If string is repeated more than once in code it should be moved to const

css-practice/task_2/video.js
```diff
@@ -0,0 +1,14 @@
+const ppbutton = document.getElementById('vidbutton');
```
tarasbobak: ppbutton what does that mean. There always should be meaningful names for everything

PR #180, opened 2018-12-03
[A.Zhukevych]Interactive-task 3

Hollister009:
https://hollister009.github.io/ui-internship-lwo2/css-practice/task3/


css-practice/task3/js/accordion.js
```diff
@@ -0,0 +1,58 @@
...
+  const panelsArr = Array.from(accordion.children);
+  const iconsArr = Array.from(accordion.querySelectorAll('.right'));
+  const imagesArr = Array.from(document.querySelectorAll('.accordion-pic'));
+
+  const switchHandler = (e) => {
+    e.preventDefault();
+    const actionBtn = e.target;
+    const iconChev = actionBtn.children[1];
+    const activePanel = actionBtn.parentNode.parentNode.parentNode;
+    const openState = activePanel.classList.contains('open') ? true : false;
```
tarasbobak: contains already return boolean :)

css-practice/task3/js/accordion.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+/* global document */
+(function() {
+  const accordion = document.getElementById('accordion');
+  const panelsArr = Array.from(accordion.children);
+  const iconsArr = Array.from(accordion.querySelectorAll('.right'));
+  const imagesArr = Array.from(document.querySelectorAll('.accordion-pic'));
+
+  const switchHandler = (e) => {
+    e.preventDefault();
```
tarasbobak: const should go first then logic

css-practice/task3/js/accordion.js
```diff
@@ -0,0 +1,58 @@
...
+      panel.classList.remove('open');
+      panel.classList.add('close');
+    });
+    iconsArr.forEach((icon) => {
+      icon.classList.remove('fa-chevron-up');
+      icon.classList.add('fa-chevron-down');
+    });
+  };
+
+  const togglePanel = (panel, state, icon) => {
```
tarasbobak: you should name openState instead of state to make thinks make more sense

css-practice/task3/js/accordion.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
```
tarasbobak: I like the way you start writing thinks and moving them to properly names blocks of code

css-practice/task3/js/accordion.js
Hollister009: true!

tarasbobak:
You're cutting the image in modal if it's too height which is bad. Secondly image should be bigger.

tarasbobak:
Generally great job.

PR #183, opened 2018-12-04
[Sofia]3rd js task(mogo template)

SofiMat:
https://sofimat.github.io/ui-internship-lwo2/html-template-3/index.html


tarasbobak:
You accordion panel has dead zone at the bottom where you click and nothing happens

tarasbobak:
You should animate accordion item height and not height of text :)
Slider looks very good.
Image viewer looks wrong for medusa image. Image should not be bigger that screen size. You need to handle that

html-template-3/script.js
```diff
@@ -0,0 +1,114 @@
...
+    accHD[i].addEventListener('click', toggleItem);
+  }
+
+  function toggleItem() {
+    const itemClass = this.parentNode.className;
+    const thisArrow = this.querySelector('.fas');
+    const currentElementIndex = accItem.indexOf(this.parentNode);
+    for (let i = 0; i < accItem.length; i++) {
+      accItem[i].className = 'drop-item close';
+      arrow[i].className = 'fas fa-angle-down';
```
tarasbobak: This class names are absolutely bad.
fas? pics?

html-template-3/script.js
```diff
@@ -0,0 +1,114 @@
...
+
+  function move() {
+    for (let i = 0; i < slide.length; i++) {
+      slide[i].style['transform'] = 'translate3d(' + offset + ', 0, 0)';
+    }
+  }
+
+  for (let i = 0; i < next.length; i++) {
+    next[i].addEventListener('click', function() {
+      if (parseInt(offset) <= -2300) {
```
tarasbobak: you are hardcoding values which is bad. I believe if you were asked to add 1 more item to slider you would hardcode another value without a meaning

html-template-3/script.js
```diff
@@ -0,0 +1,114 @@
...
+
+  const navigation = document.querySelector('.top-navigation');
+  const menuItem = document.getElementsByClassName('menu-item');
+
+  navigation.addEventListener('click', (e) => addBorder(e.target.id));
+
+  function addBorder(id) {
+    const selectedLink = document.getElementById(`${id}`);
+    for (let i = 1; i <= menuItem.length; i++) {
+      const link = document.getElementById(`${i}`);
```
tarasbobak: again, id's. You should use classes instead of id's 1,2,3.

html-template-3/script.js
```diff
@@ -0,0 +1,114 @@
...
+(function slider() {
+  'use strict';
+  const next = [...document.querySelectorAll('.fa-angle-right')];
+  const prev = [...document.querySelectorAll('.fa-angle-left')];
+  const slide = [...document.querySelectorAll('.quote-slider')];
+  let offset = 0;
+
+  function move() {
+    for (let i = 0; i < slide.length; i++) {
+      slide[i].style['transform'] = 'translate3d(' + offset + ', 0, 0)';
```
tarasbobak: If you are moving by X axis only you should use translateX

tarasbobak:
Code is not very good. Think more about naming.

PR #189, opened 2018-12-05
[stas-molchanov]js practice task 3

moskdev:
https://moskdev.github.io/ui-internship-lwo2/css-practice/task-3/


css-practice/task-3/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+      el.querySelector('i').classList.remove('rotate-180');
+    }
+  });
+
+  e.currentTarget.querySelector('i').classList.toggle('rotate-180');
+
+  if (e.currentTarget.nextElementSibling.classList.contains('height-0')) {
+    changeAccordionImage(e.currentTarget.getAttribute('data-image'));
+    servicesAccordionText.forEach((el) => {
+      el.classList.remove('height-auto');
```
tarasbobak: very strange names for classes.
You should be describing state and not the styles which you apply
So it should be collapse and expanded

css-practice/task-3/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+
+function carouselPrev(e) {
+  let active = e.currentTarget.parentElement.querySelector('.active-carou-item');
+
+  if (!active.previousElementSibling) {
+    active.classList.remove('active-carou-item');
+    active.classList.add('prev-carou-item');
+    active.nextElementSibling.classList.remove('next-carou-item');
+    active.nextElementSibling.classList.add('prev-carou-item');
+    active.nextElementSibling.nextElementSibling.classList.remove('next-carou-item');
```
tarasbobak: It is dumb implementation :) what would you do if you were asked to add 1 more item to carousel?)

tarasbobak: . nextElementSibling. nextElementSibling. nextElementSibling?

css-practice/task-3/js/script.js
```diff
@@ -0,0 +1,113 @@
...
+
+function carouselNext(e) {
+  let active = e.currentTarget.parentElement.querySelector('.active-carou-item');
+
+  if (!active.nextElementSibling) {
+    active.classList.remove('active-carou-item');
+    active.classList.add('next-carou-item');
+    active.previousElementSibling.classList.remove('prev-carou-item');
+    active.previousElementSibling.classList.add('next-carou-item');
+    active.previousElementSibling.previousElementSibling.classList.remove('prev-carou-item');
```
tarasbobak: same as below

tarasbobak:
Visually looks pretty good.
You should always add cursor: pointer if there are button you can click, like prev and next in carousel

css-practice/task-3/js/script.js
moskdev: https://developer.mozilla.org/ru/docs/Web/API/Node/firstChild
i wish i knew about this earlier

PR #191, opened 2018-12-05
[A.Zhukevych]Module-pattern

Hollister009:
https://hollister009.github.io/ui-internship-lwo2/css-practice/cooperative-task/


tarasbobak:
Good

PR #192, opened 2018-12-05
[TARAS YAKYMIUK] Module pattern first task

taras1313:


css-practice/task2/workspace/form-validator.js
```diff
@@ -1,4 +1,5 @@
 /* global document */
+// const formValidator = (() => {
```
tarasbobak: Why do you forget to remove these stupid comments?

tarasbobak:
Good job

PR #193, opened 2018-12-05
[Natalia Falinska] Module pattern

NFalinska:


css-practice/task_2/video-module.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable max-len */
+/* global  document */
+const video = document.getElementById('my-video');
+const btnVideo = document.getElementById('play');
+
+function togglePlay() {
+        video.paused ? playVideo() : pauseVideo();
```
tarasbobak: Always indent properly. By this time you should not do this kind of mistakes

css-practice/task_2/main.js
```diff
@@ -0,0 +1,6 @@
+import videoPlay from './video-module.js';
+import formValid from './form-module.js';
+
+
+formValid.valid();
```
tarasbobak: why valid and not init?

css-practice/task_2/form-module.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable max-len */
+/* global  document */
+const loginFormInpunts = document.forms['main-form'].getElementsByTagName('input');
+
+function validate(event) {
+  if (!event.target.validity.valid || !event.target.value) {
+    event.target.classList.add('invalid');
```
tarasbobak: every string used more than once in code has to be put in const

tarasbobak:
Good job

PR #194, opened 2018-12-05
[Sofia]Module pattern advanced

SofiMat:


css-practice/task2/formValidationModule.js
```diff
@@ -0,0 +1,96 @@
...
+disable();
+
+const formValidationModuleApi = {
+  init: () => {
+    handleFormValidation(['contact-form']);
+    FORMS_ON_PAGE.forEach((el) => {
+      el.addEventListener('submit', submit);
+    });
+
+    function handleFormValidation(formName) {
```
tarasbobak: handleFormValidation can be put on the same level as submit, disable functions

css-practice/task2/formValidationModule.js
```diff
@@ -0,0 +1,96 @@
...
+
+const PATTERNS = {
+  name: /[A-Za-z]{2,20}/g,
+  /* eslint-disable-next-line */
+  email: /^([,a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9_\-\.]+)\.([a-zA-Z]{2,5})$/g,
+  /* eslint-disable-next-line */
+  password: /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{6,}$/g,
+};
+
+const FORMS_ON_PAGE = document.querySelectorAll('form');
```
tarasbobak: THIS_PATTERN was used in es5 to notify other dev this var should not be changed.
Nowadays we use this pattern rarely. It is almost always enough to type const

PR #199, opened 2018-12-07
[Pavlo Turchyn] Module pattern

FabianMoretti:


tarasbobak:
Why do I still see 42 files changed???

PR #204, opened 2018-12-09
[Natalia Falinska] fetch for task3

NFalinska:


css-practice/task_3/index.js
```diff
@@ -0,0 +1,48 @@
+/* global document */
+const blogs = document.querySelector('.stories .pics');
+
+function render(previewImg, day, month, title, watched, comments, description) {
```
dverbovyi: render should accept a response object

and all parsing, looping logic should be here, and not if fetchData

css-practice/task_3/index.js
```diff
@@ -0,0 +1,48 @@
...
+    `;
+
+  element.innerHTML = blog;
+  blogs.append(element);
+}
+
+(function fetchData() {
+  fetch('http://localhost:3000/api/blogs').then((res) => res.json()).then((res) => {
+    const result = res.blogs;
+    for (let i = 0; i < result.length; i++) {
```
dverbovyi: move this looping into render

css-practice/task_3/latest.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable max-len */
+/* global document */
+const latests = document.querySelector('.blocks .block-two');
```
dverbovyi: global scope variable detected! Move to the place where it's using in the local function

css-practice/task_3/latest.js
```diff
@@ -0,0 +1,32 @@
...
+     <div class="date">${published}</div>
+  </div>
+    `;
+
+  element.innerHTML = latest;
+  latests.append(element);
+}
+
+(function fetchData() {
+  fetch('http://localhost:3000/api/blogs').then((res) => res.json()).then((res) => {
```
dverbovyi: the second request detected!

You should send the only 1 request.

css-practice/task_3/latest.js
```diff
@@ -0,0 +1,32 @@
...
+    const result = res.blogs;
+    for (let i = 0; i < result.length; i++) {
+      const curDate = new Date(result[i].published);
+      renderLatest(result[i].previewImg,
```
dverbovyi: you're rendering all blogs in the latest section. Where is filtering by ids in res.latest ?

css-practice/task_3/index.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable max-len */
+/* global document */
+function render(res) {
```
dverbovyi: this function is a trash-bin. The function should do only 1 thing as Robert Martin said. (You should know about it If you truly have read his book)
This function is doing everything.
Filtering, date formatting and rendering both blogs sections (latest blogs and footer ones).
All of this should be logical separated

dverbovyi: It's very hard to read

PR #207, opened 2018-12-10
[Y_Demkiv] Task 4 blogs fetch

YarynaD:


css-practice/task_3/blog.js
```diff
@@ -0,0 +1,73 @@
...
+  res[1].filter(function(item, index) {
+    return res[0].includes(index);
+}).forEach(((model)=> {
+    let tempDate = new Date(model.published);
+    let locale = 'en-us';
+    let month = tempDate.toLocaleString(locale, {month: 'short'});
+    let theDate = [tempDate.getFullYear(), month, tempDate.getDate()];
+    eventsSection.appendChild(new RenderBlogItem(model, theDate));
+  }));
+  res[1].filter(function(item, index) {
```
dverbovyi: duplicated code found. Please don't repeat yourself

- At first, create a separate parseDate function, which will return an object with {month, day, year} fields

- .forEach loop is completely the same, move to a common function which will receive section to append and render specific Item function.

- don't use new for this types of function. It's not class, just call them
renderBlogItem(model)

css-practice/task_3/blog.js
```diff
@@ -0,0 +1,73 @@
+const eventsSection = document.querySelector('.events-list');
+const footerSection = document.querySelector('.footer-blogs');
```
dverbovyi: global scope variables detected!

PR #212, opened 2018-12-10
[Sofia]Fetch task

SofiMat:


html-template-3/export.js
```diff
@@ -0,0 +1,38 @@
...
+
+function fetchData() {
+  return fetch('http://localhost:3000/api/blogs')
+      .then((response) => {
+        return response.json();
+      })
+      .then(dataFilter);
+}
+
+export const formatDate = (date) => {
```
dverbovyi: move this function to blog.js because it's not using here

html-template-3/export.js
```diff
@@ -0,0 +1,38 @@
+/* global fetch */
```
dverbovyi: give a more appropriate name for this file

html-template-3/export.js
SofiMat: done

html-template-3/export.js
SofiMat: done

PR #215, opened 2018-12-18
[TARAS YAKYMIUK]Prepushhooks

taras1313:
the problem is, that i cannot run ng-test using phantomJS, so, as Dmytro, i use in this version chrome for karma.


package.json
```diff
@@ -20,6 +26,7 @@
     "@angular/platform-browser-dynamic": "~7.1.0",
     "@angular/router": "~7.1.0",
     "core-js": "^2.5.4",
+    "intl": "^1.2.5",
```
dverbovyi: why do you need this?

